Issue #31, opened 2024-08-19
Handle the migration job for PostgreSQL-backed servers


chrisguidry:
In an effort to support multiple replicas of the Prefect server (a configuration that is largely possible when using a PostgreSQL backend), we would need to run the migration command `prefect server database migrate --yes` in a `Job` whenever the version or the database credentials changed.  We shouldn't depend on the Prefect server running migrations at startup, because the multiple servers running concurrently will be stepping on one another as they do so, leading to lots of additional errors in the logs and possibly data consistency issues if the migrations aren't designed with idempotency in mind.

There is a simple version of this in the current controller that will create the `Job` the _first_ time the `PrefectServer` is created, but we don't yet have the logic for recreating the Job when other `PrefectServer` options have changed.  The current complication is that `Job`s are largely read-only after creation, and we'd need to create a _new_ job each time.

One way to approach this would be to create a `Job` with each Prefect `major.minor.patch` version, so that migrations are applied during upgrades.  This may not help with reconciling other server-related changes (like database connection info, etc), but it would guarantee that migrations were applied when necessary.

PR #45, opened 2024-08-27
feat: add update reconciliation for migration job

parkedwards:
resolves https://github.com/PrefectHQ/prefect-operator/issues/31

migration `Job`s will be handled differently in that anytime a new `Job` spec is generated, we'll create a new, distinct `Job` resource.  This append-only operation allows us to not worry about Deletes, as `Job` resources (usually) need to be replaced if there are updates to its spec

### DIFF
--- a/internal/controller/prefectserver_controller.go
+++ b/internal/controller/prefectserver_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"dario.cat/mergo"
 	appsv1 "k8s.io/api/apps/v1"
@@ -207,7 +208,9 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
 	case !isMigrationJobFinished(foundMigrationJob):
 		log.Info("Waiting on active migration Job to complete", "name", foundMigrationJob.Name)
 		condition = conditions.AlreadyExists(objName, fmt.Sprintf("migration Job %s is still active", foundMigrationJob.Name))
-		return &ctrl.Result{Requeue: true}, nil
+
+		// We'll requeue after 20 seconds to check on the migration Job's status
+		return &ctrl.Result{Requeue: true, RequeueAfter: 20 * time.Second}, nil
 
 	default:
 		if !meta.IsStatusConditionTrue(server.Status.Conditions, "MigrationJobReconciled") {
--- a/internal/controller/prefectserver_controller.go
+++ b/internal/controller/prefectserver_controller.go
@@ -204,13 +204,20 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
 	} else if migrationJobNeedsUpdate(&foundMigrationJob.Spec, &desiredMigrationJob.Spec, log) {
 		log.Info("Recreating migration Job", "name", desiredMigrationJob.Name)
 		condition = conditions.NeedsUpdate(objName)
-		if err = r.Update(ctx, desiredMigrationJob); err != nil {
-			condition = conditions.UpdateFailed(objName, err)
 
+		// deletes the job + recreates it.
+		// k8s Jobs are immutable and need to be recreated if they are to be re-run.
+		if err = r.Delete(ctx, foundMigrationJob); err != nil {
+			condition = conditions.NotDeleted(objName, err)
 			return &ctrl.Result{}, err
 		}
 
-		condition = conditions.Updated(objName)
+		if err = r.Create(ctx, desiredMigrationJob); err != nil {
+			condition = conditions.NotCreated(objName, err)
+			return &ctrl.Result{}, err
+		}
+
+		condition = conditions.Created(objName)
 	} else {
 		if !meta.IsStatusConditionTrue(server.Status.Conditions, "MigrationJobReconciled") {
 			condition = conditions.Updated(objName)
--- a/internal/controller/prefectserver_controller.go
+++ b/internal/controller/prefectserver_controller.go
@@ -201,9 +201,35 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
 		condition = conditions.AlreadyExists(objName, errorMessage)
 
 		return &ctrl.Result{}, errors.NewBadRequest(errorMessage)
-	} else if jobNeedsUpdate(&foundMigrationJob.Spec, &desiredMigrationJob.Spec, log) {
-		// TODO: handle replacing the job if something has changed
+	} else if migrationJobNeedsUpdate(&foundMigrationJob.Spec, &desiredMigrationJob.Spec, log) {
+		log.Info("Recreating migration Job", "name", desiredMigrationJob.Name)
+		// deletes the job + recreates it.
+		// k8s Jobs are immutable and need to be recreated if they are to be re-run.
+		if err = r.Delete(ctx, foundMigrationJob); err != nil {
+			if statusErr := r.updateCondition(ctx, server, metav1.Condition{
+				Type:    "MigrationJobReconciled",
+				Status:  metav1.ConditionFalse,
+				Reason:  "MigrationJobNotDeleted",
+				Message: "MigrationJob was not deleted: " + err.Error(),
+			}); statusErr != nil {
+				return &ctrl.Result{}, statusErr
+			}
 
+			return &ctrl.Result{}, err
+		}
+
+		if err = r.Create(ctx, desiredMigrationJob); err != nil {
+			if statusErr := r.updateCondition(ctx, server, metav1.Condition{
+				Type:    "MigrationJobReconciled",
+				Status:  metav1.ConditionFalse,
+				Reason:  "MigrationJobNotRecreated",
+				Message: "MigrationJob was not recreated: " + err.Error(),
+			}); statusErr != nil {
+				return &ctrl.Result{}, statusErr
+			}
+
+			return &ctrl.Result{}, err
+		}
 	} else {
 		if !meta.IsStatusConditionTrue(server.Status.Conditions, "MigrationJobReconciled") {
 			condition = conditions.Updated(objName)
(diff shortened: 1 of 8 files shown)


internal/controller/prefectserver_controller.go
```diff
@@ -363,9 +389,9 @@ func pvcNeedsUpdate(current, desired *corev1.PersistentVolumeClaimSpec, log logr
 	return false
 }
 
-func jobNeedsUpdate(current, desired *batchv1.JobSpec, log logr.Logger) bool {
-	// TODO: check for changes to the job spec that require an update
-	return false
+func migrationJobNeedsUpdate(current, desired *batchv1.JobSpec, log logr.Logger) bool {
+	merged := current.DeepCopy()
+	return needsUpdate(current, merged, desired, log)
```
parkedwards, line 394: originally i thought to stash the prefect version / postgres connection info in a configmap and do an update check on that resource to determine if the migration job should be re-created

but we store that info in the Job spec already, in the image (which is computed from the CR version) and the env vars (where we pass in the CR's postgres connection parameters).   

im thinking we just do the same deepmerge check on the Job spec itself since it contains all of the version/db connection information, and if any of that changes we can recreate the Job

internal/controller/prefectserver_controller.go
```diff
@@ -203,9 +204,23 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
...
+			log.Info("Deleting existing migration Job", "name", foundMigrationJob.Name)
+			if err = r.Delete(ctx, foundMigrationJob); err != nil {
+				condition = conditions.NotDeleted(objName, err)
+				return &ctrl.Result{}, err
+			}
+			condition = conditions.Deleted(objName)
+		} else {
+			log.Info("Waiting for migration Job to be deleted, so it can be recreated", "name", desiredMigrationJob.Name)
+		}
+		return &ctrl.Result{Requeue: true}, nil
```
parkedwards, line 223: there's no "waiting" in the operator pattern, so we have to make this reconciliation handler idempotent. so:
- if we detect that an update (actually, a replacement) is needed
- we'll take the existing Job resource, see if it's been marked for deletion -> if not, delete it
- we can't be sure of how many reconciliations it will take, so we'll assume it takes more than 1
- we'll continue to requeue the reconciliation job until the `foundMigrationJob` is cleaned up, so that the very next run will create a new Job

parkedwards: https://sdk.operatorframework.io/docs/faqs/#how-do-i-wait-for-some-specific-cluster-state-such-as-a-resource-being-deleted

internal/controller/prefectserver_controller.go
```diff
@@ -203,8 +204,24 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
...
+	} else if migrationJobNeedsUpdate(&foundMigrationJob.Spec, &desiredMigrationJob.Spec, log) {
+		log.Info("Will replace migration Job", "name", desiredMigrationJob.Name)
+		condition = conditions.NeedsUpdate(objName)
+
+		// deletes the Job, so it can be recreated on the following reconciliation.
+		// k8s Jobs are immutable and need to be recreated if they are to be re-run.
+		if foundMigrationJob.GetDeletionTimestamp() == nil {
+			log.Info("Deleting existing migration Job", "name", foundMigrationJob.Name)
+
+			if err = r.Delete(ctx, foundMigrationJob); err != nil {
```
parkedwards, line 216: for some odd reason, our test cases never see this `Job` resource get deleted (only soft deleted), so a subsequent create will always fail 🤔

internal/controller/prefectserver_controller.go
```diff
@@ -627,30 +639,35 @@ func (r *PrefectServerReconciler) postgresDeploymentSpec(server *prefectiov1.Pre
...
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{
-						{
-							Name:    "prefect-server-migration",
-							Image:   server.Image(),
-							Command: []string{"prefect", "server", "database", "upgrade", "--yes"},
-							Env:     append(append(server.ToEnvVars(), server.Spec.Postgres.ToEnvVars()...), server.Spec.Settings...),
-						},
+	jobSpec := batchv1.JobSpec{
+		TTLSecondsAfterFinished: ptr.To(int32(7 * 24 * 60 * 60)), // 7 days
```
parkedwards, line 643: since we're creating a new Job on every PrefectServer update, we'll have Jobs like this:

```
kubectl get jobs

prefect-migration-1234
prefect-migration-5678
prefect-migration-2956
...
```

we'll add a 1-week TTL that cleans up the Job after it is complete - Im thinking this gives enough of a window for retrieving past Job runs and no-op-ing if the Job Spec hasn't changed

internal/controller/prefectserver_controller.go
```diff
@@ -179,32 +182,34 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
 
 	foundMigrationJob := &batchv1.Job{}
 	err = r.Get(ctx, types.NamespacedName{Namespace: server.Namespace, Name: desiredMigrationJob.Name}, foundMigrationJob)
-	if errors.IsNotFound(err) {
+
+	switch {
+	case errors.IsNotFound(err):
```
parkedwards, line 187: i like to use switch cases when the if/else branching gets > 3

internal/controller/prefectserver_controller_test.go
```diff
@@ -1548,6 +1566,18 @@ var _ = Describe("PrefectServer controller", func() {
...
+				_, _, desiredMigrationJob = controllerReconciler.prefectServerDeployment(prefectserver)
+				migrateJob = &batchv1.Job{}
+				Eventually(func() error {
+					return k8sClient.Get(ctx, types.NamespacedName{
+						Namespace: namespaceName,
+						Name:      desiredMigrationJob.Name,
+					}, migrateJob)
+				}).Should(Succeed())
+				migrateJob.Status.Succeeded = 1
+				Expect(k8sClient.Status().Update(ctx, migrateJob)).To(Succeed())
```
parkedwards, line 1580: kind of annoying, but looking up a `Job` by name means i have to re-run the manifest generator every time to create the hashed name, which is based on the manifest

mitchnielsen: Yeah, I remember thinking about using `.metadata.generation` from the CR since that should be unique and change each time the CR changes. But not sure if there are any drawbacks to that approach.

internal/utils/hash_test.go
```diff
@@ -0,0 +1,67 @@
...
+		By("comparing hash for different job specs")
+		job := MigrationJobStub()
+		hash1, err := Hash(job.Spec, 8)
+		Expect(err).NotTo(HaveOccurred())
+
+		job.Spec.Template.Spec.Containers[0].Image = "image1:v2"
+		hash2, err := Hash(job.Spec, 8)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(hash1).NotTo(Equal(hash2))
```
mitchnielsen, line 65: noice

internal/controller/prefectserver_controller.go
```diff
@@ -627,30 +638,35 @@ func (r *PrefectServerReconciler) postgresDeploymentSpec(server *prefectiov1.Pre
...
 			},
 		},
 	}
+
+	// Generate hash based on the job spec
+	hashSuffix, _ := utils.Hash(jobSpec, 8) // Use the first 8 characters of the hash
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: server.Namespace,
+			Name:      fmt.Sprintf("%s-migration-%s", server.Name, hashSuffix),
```
mitchnielsen, line 666: We may eventually want to trim this name (and the name of all resources, I guess) to 63 characters in case the `server.Name` provided is unusually long for some reason. Or we can let it fail and we'll know why - just calling it out as a familiar speedbump.

internal/controller/prefectserver_controller.go
```diff
@@ -179,32 +182,34 @@ func (r *PrefectServerReconciler) reconcileMigrationJob(ctx context.Context, ser
...
-
 		return &ctrl.Result{}, errors.NewBadRequest(errorMessage)
-	} else if jobNeedsUpdate(&foundMigrationJob.Spec, &desiredMigrationJob.Spec, log) {
-		// TODO: handle replacing the job if something has changed
 
-	} else {
+	case !isMigrationJobFinished(foundMigrationJob):
+		log.Info("Waiting on active migration Job to complete", "name", foundMigrationJob.Name)
+		condition = conditions.AlreadyExists(objName, fmt.Sprintf("migration Job %s is still active", foundMigrationJob.Name))
+		return &ctrl.Result{Requeue: true}, nil
```
mitchnielsen, line 210: From memory, this will immediately requeue the reconcile loop. If we find that it ends up looping a ton while waiting for these to finish (I'm not sure how long they usually take), we can pass another field like this:

```go
return &ctrl.Result{Requeue,: true, RequeueAfter: 60 * time.Seconds}
```

mitchnielsen (approved):
nice! looks good

internal/controller/prefectserver_controller.go
chrisguidry: I forgot that in Go you can use full expressions in the switch cases :bow:

internal/controller/prefectserver_controller.go
chrisguidry: Perfect!

chrisguidry (approved):
Dig it!  This will get even cleaner when we give it the `CreateOrUpdate` treatment in a future PR

internal/controller/prefectserver_controller.go
parkedwards: ah - good callout.  it's probably a good idea to add a time buffer in so it doesn't spam the loop

internal/controller/prefectserver_controller.go
parkedwards: good call out. might make sense to validate a char limit for the .Name on the CR
https://github.com/PrefectHQ/prefect-operator/issues/56

PR #47, opened 2024-08-27
Remove scaffolding specific to operator-sdk

mitchnielsen:
## Summary

Removes the `operator-sdk`-specific scaffolding, instead relying only on the scaffolding from `kubebuilder`.

We don't currently have plans to push to ArtifactHub or OLM, so let's remove it for now.

## References

- https://book.kubebuilder.io/cronjob-tutorial/basic-project.html
- https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project
- https://sdk.operatorframework.io/docs/overview/project-layout

PR #48, opened 2024-08-28
Restores CRD descriptions

chrisguidry:
In our [previous change to de-scaffold the extra operator-sdk boilerplate](https://github.com/PrefectHQ/prefect-operator/pull/47), we
lost the descriptions in our CRDs.  Since we're not experiencing the issue
described in the comment, we can keep our descriptions for now to help with
documentation.


mitchnielsen (approved):
Works for me! I saw this change had made it into `main` in the template scaffolding, and I remember hitting this issue in the past, but definitely makes sense to keep it around while we can.

PR #50, opened 2024-08-28
Adds `Ready` status boolean for `PrefectServer` and `PrefectWorkPool`

### LINKED ISSUE #49: Readiness status on `PrefectServer`
The `PrefectServer` should have a status attribute indicating when it is available to receive requests.  This should be derived from the underlying deployment and should only be updated when there is at least 1 `availableReplicas`

chrisguidry:
For both of these, `Ready` will be true when there is at least one ready
replica of the underlying deployment.  While working on this, I also noticed
that we were _always_ attempting to add a condition, even if nothing had
changed (and thus the condition struct was empty), which would lead to quiet
errors with all further reconciliations.  Now we'll exit early if the condition
isn't populated, report errors if we can't update the status, and explicitly
update the status when non-conditions are changed.

Fixes #49

### DIFF
--- a/internal/conditions/conditions.go
+++ b/internal/conditions/conditions.go
@@ -6,82 +6,73 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func newCondition(typeVal, reason, message string, status metav1.ConditionStatus) metav1.Condition {
+func NotRequired(object string) metav1.Condition {
 	return metav1.Condition{
-		Type:    typeVal,
-		Status:  status,
-		Reason:  reason,
-		Message: message,
+		Type:    fmt.Sprintf("%sReconciled", object),
+		Reason:  fmt.Sprintf("%sNotRequired", object),
+		Message: fmt.Sprintf("%s is not required", object),
+		Status:  metav1.ConditionTrue,
 	}
 }
 
-func NotRequired(object string) metav1.Condition {
-	return newCondition(
-		fmt.Sprintf("%sReconciled", object),
-		fmt.Sprintf("%sNotRequired", object),
-		fmt.Sprintf("%s is not required", object),
-		metav1.ConditionTrue,
-	)
-}
-
 func NotCreated(object string, err error) metav1.Condition {
-	return newCondition(
-		fmt.Sprintf("%sReconciled", object),
-		fmt.Sprintf("%sNotCreated", object),
-		fmt.Sprintf("%s was not created: %v", object, err.Error()),
-		metav1.ConditionFalse,
-	)
+	return metav1.Condition{
+		Type:    fmt.Sprintf("%sReconciled", object),
+		Reason:  fmt.Sprintf("%sNotCreated", object),
+		Message: fmt.Sprintf("%s was not created: %v", object, err.Error()),
+		Status:  metav1.ConditionFalse,
+	}
 }
 
 func Created(object string) metav1.Condition {
-	return newCondition(
-		fmt.Sprintf("%sReconciled", object),
-		fmt.Sprintf("%sCreated", object),
-		fmt.Sprintf("%s was created", object),
-		metav1.ConditionTrue,
-	)
+	return metav1.Condition{
+		Type:    fmt.Sprintf("%sReconciled", object),
+		Reason:  fmt.Sprintf("%sCreated", object),
+		Message: fmt.Sprintf("%s was created", object),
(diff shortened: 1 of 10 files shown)


internal/conditions/conditions.go
```diff
@@ -6,82 +6,73 @@ import (
...
+func NotRequired(object string) metav1.Condition {
 	return metav1.Condition{
-		Type:    typeVal,
-		Status:  status,
-		Reason:  reason,
-		Message: message,
+		Type:    fmt.Sprintf("%sReconciled", object),
+		Reason:  fmt.Sprintf("%sNotRequired", object),
+		Message: fmt.Sprintf("%s is not required", object),
+		Status:  metav1.ConditionTrue,
```
chrisguidry, line 14: While I was debugging conditions, I realized that the `newCondition` function wasn't earning us much, so I just inlined the conditions below to make it more clear what is used for which value.

internal/controller/prefectserver_controller.go
```diff
@@ -270,20 +267,26 @@ func (r *PrefectServerReconciler) reconcileDeployment(ctx context.Context, serve
...
+		version := prefectiov1.VersionFromImage(desiredDeployment.Spec.Template.Spec.Containers[0].Image)
 
-			condition = conditions.Updated(objName)
+		if server.Status.Ready != ready || server.Status.Version != version {
+			server.Status.Ready = ready
+			server.Status.Version = version
+
+			if statusErr := r.Status().Update(ctx, server); statusErr != nil {
+				return &ctrl.Result{}, statusErr
+			}
```
chrisguidry, line 283: I decided to uncouple the condition update from the update of any other status fields, because it was hard to determine if the status needed changing separately from the conditions.  This split makes it much more explicitly when we're updating the status.

internal/conditions/conditions.go
mitchnielsen: Yeah I wavered on that change, works for me 👍🏼

internal/controller/prefectserver_controller.go
```diff
@@ -270,20 +267,26 @@ func (r *PrefectServerReconciler) reconcileDeployment(ctx context.Context, serve
...
 	} else {
-		imageVersion := prefectiov1.VersionFromImage(desiredDeployment.Spec.Template.Spec.Containers[0].Image)
-		if server.Status.Version != imageVersion ||
-			!meta.IsStatusConditionTrue(server.Status.Conditions, "DeploymentReconciled") {
+		if !meta.IsStatusConditionTrue(server.Status.Conditions, "DeploymentReconciled") {
+			condition = conditions.Updated(objName)
+		}
 
-			server.Status.Version = imageVersion
+		ready := foundDeployment.Status.AvailableReplicas > 0
```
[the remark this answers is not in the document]
mitchnielsen: For example, [here is how the Kubernetes Deployment controller calculates if a Deployment is complete](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L706-L713):

```go
// DeploymentComplete considers a deployment to be complete once all of its desired replicas
// are updated and available, and no old pods are running.
func DeploymentComplete(deployment *apps.Deployment, newStatus *apps.DeploymentStatus) bool {
	return newStatus.UpdatedReplicas == *(deployment.Spec.Replicas) &&
		newStatus.Replicas == *(deployment.Spec.Replicas) &&
		newStatus.AvailableReplicas == *(deployment.Spec.Replicas) &&
		newStatus.ObservedGeneration >= deployment.Generation
}
```

mitchnielsen (approved):

internal/controller/prefectserver_controller.go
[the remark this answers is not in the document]
chrisguidry: Great question!  What I was hoping for here is a simple boolean signal telling us whether the `PrefectServer` is ready to accept traffic, which would be the case if there's at least 1 available replica.  During an upgrade, we might go up to 2 and then back down to 1, and during that time we're still available and ready to receive requests.  I think that's what we want here.

But now you've got me thinking about how the `Status.Version` fields work, which is looking at the Deployment and not the underlying Pods.  The truth is that you could have multiple versions running in parallel during an upgrade, so maybe that should be a list like:

```
status: [
  {
    version: "3.1.2",
    ready: 2
  },
  {
    version: "3.1.3",
    ready: 1
  },
]
```

This might be useful to understand during a version bump, not sure yet...

PR #61, opened 2024-09-03
Support specifying which namespace(s) to watch

mitchnielsen:
By default, an Operator is configured to watch objects from any namespace. Its access is only limited by RBAC.

The Prefect Operator should only need to manage resources in the same namespace where it is deployed. So let's default to restricting it to its own namespace.

References:
- https://book.kubebuilder.io/cronjob-tutorial/empty-main.html
- https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope


mitchnielsen:
Wanted to throw this up as a `Draft` for later in case this ends up making sense for this Operator. As a note, if we pursue this, we'll want to [update RBAC](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#restricting-roles-and-permissions) to match the new permission model (i.e. change `ClusterRole` to `Role`, etc).

chrisguidry:
> The Prefect Operator should only need to manage resources in the same namespace where it is deployed.

I disagree with this statement, the idea is to run a single `prefect-operator` to deploy prefect servers and workers in any number of namespaces within a cluster.

Would this change support both approaches?  Is it just additive or does this remove the ability to have a single operator manage a whole cluster?

mitchnielsen:
> the idea is to run a single prefect-operator to deploy prefect servers and workers in any number of namespaces within a cluster

Ah, good to know. That should be fine - my only concern is the version skew. If some customers want the latest, and some customers want/need to stay on fairly old versions, the Operator has to support that version spread. But happy to proceed as-is for now.

> Would this change support both approaches? Is it just additive or does this remove the ability to have a single operator manage a whole cluster?

This is just additive since we control the `WATCH_NAMESPACES` environment variable. If it's not set at all, it'll be able to monitor all namespaces.

I thought of this when we were considering testing this in existing cluster(s) and wanted to have the option to maintain isolation if possible.

chrisguidry:
Sweet, if this is just additive, I'm great with it!  I just didn't want to cut off the ability to have a simple cluster-wide operator.

I'm not too too worried about version skew for Prefect itself, because the deployment topology for a Prefect server has been roughly the same since 2.0.  There have been a few tweaks (like the change in 3.0 that allows you to specify all the individual components of a DB connection as separate envvars), but I don't expect things to change much until Prefect has more built-in redis support (and even that isn't a huge change deployment-wise)

config/manager/manager.yaml
```diff
@@ -62,6 +62,13 @@ spec:
         - /manager
         args:
         - --leader-elect
+        env:
+        - name: WATCH_NAMESPACE
+          # This restricts the Operator to watch objects in the same
+          # namespace where it is deployed.
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
chrisguidry, line 71: I think we will want to have two versions of the operator manifest, or just have this snippet be mentioned in the docs for how to constrain it to its own namespace.

config/manager/manager.yaml
```diff
@@ -62,6 +62,13 @@ spec:
         - /manager
         args:
         - --leader-elect
+        env:
+        - name: WATCH_NAMESPACE
```
chrisguidry, line 66:
```suggestion
        - name: WATCH_NAMESPACES
```

right?

config/manager/manager.yaml
mitchnielsen: Yeah, the `operator-sdk` docs are the only place that mention how to make the permissions dynamic - but it only covers the Golang part, not the RBAC manifests part. So we can have two, or just a `kustomize` command to inject it, or leave it commented out and just hard-code it when needed. Will tinker with it.

mitchnielsen: Opened https://github.com/PrefectHQ/prefect-operator/issues/72 to track this.

chrisguidry (approved):

PR #62, opened 2024-09-03
Refactors controllers to improve test coverage

chrisguidry:
With these changes, we're at 57% coverage (up from 55%), all while covering
more modules than before.  There were still some redundant status API calls
that would be caught with the outer deferred final update, as well as some
unused conditions.  We had a nice utility for picking the right condition based
on an operation result and error, so I'm using that throughout.


internal/conditions/conditions.go
```diff
@@ -4,8 +4,25 @@ import (
 	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+func GetStatusConditionForOperationResult(result controllerutil.OperationResult, objectName string, err error) metav1.Condition {
```
parkedwards, line 10: nice

chrisguidry: All credit to @mitchnielsen, I just moved it here to avoid an extra file

parkedwards (approved):

PR #63, opened 2024-09-03
Adds CODEOWNERS

chrisguidry:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*                                    @chrisguidry @parkedwards @mitchnielsen
```
mitchnielsen, line 1: Is there a lot of spacing between `*` and `@chris...` or is that just on my end?

mitchnielsen (approved):

.github/CODEOWNERS
chrisguidry: Ha, I stole this from Prefect's top-level one and it had a lot of reserved space for other paths.  Totally not necessary here, will fix

Issue #66, opened 2024-09-05
Fix image pushes to Docker Hub production repository


mitchnielsen:
## Summary

In [https://github.com/PrefectHQ/prefect-operator/actions/runs/10729548105/job/29756401909](<https://github.com/PrefectHQ/prefect-operator/actions/runs/10729548105/job/29756401909>), I'm getting an error:

```
Error: buildx failed with: ERROR: failed to solve: failed to push prefecthq/prefect-operator:main: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

mitchnielsen:
<img width="738" alt="image" src="https://github.com/user-attachments/assets/900abf3b-ebf7-4c5c-ab3b-752415ac1ea7">

Looks like the team settings are correct. Wonder if the `production` GitHub Secrets aren't being selected.

mitchnielsen:
Yeah, my GH Actions logic needs to change from checking for a release to checking for a tag.

mitchnielsen:
Alright, closer now - but in [this run](https://github.com/PrefectHQ/prefect-operator/actions/runs/10730591043/job/29759330078) on a tag, it still picked the `dev` repository 🤔

mitchnielsen:
Assets got uploaded though 👍🏼 https://github.com/PrefectHQ/prefect-operator/releases (at the bottom)

mitchnielsen:
All set now 🎉

PR #68, opened 2024-09-06
Fix production docker image name

mitchnielsen:
Returning '' is considered null in GitHub's action evaluation, which was causing it to always return the '-dev'-suffixed image name.

This revamps the image name calculation so it returns a valid string in both cases.

Related to https://github.com/PrefectHQ/prefect-operator/issues/66


parkedwards (approved):

parkedwards:
You're discovering the joy of trying to use logic in yaml

mitchnielsen:
image

PR #69, opened 2024-09-06
Align the default server name in the sample

mitchnielsen:
## Summary

The sample configuration we enable by default picks prefect-sqlite as the server, but the sample work pool referred to prefect-ephemeral.

We can still change this around of course if we want a different default sample, but this is what's currently configured in kustomization.yaml.

## Testing

```yaml
$ kustomize build config/samples
apiVersion: prefect.io/v1
kind: PrefectServer
metadata:
  labels:
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: prefect-operator
  name: prefect-sqlite # <== matching
spec:
  sqlite:
    size: 1Gi
    storageClassName: standard
---
apiVersion: prefect.io/v1
kind: PrefectWorkPool
metadata:
  labels:
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: prefect-operator
  name: sample-pool
spec:
  server:
    name: prefect-sqlite # <== matching
  type: process
  workers: 3
```


chrisguidry (approved):
Ah good catch, thank you!

PR #70, opened 2024-09-06
Allows work pools for Prefect servers without API keys, and includes examples

chrisguidry:
We were inadvertently requiring an API key to connect to a remote Prefect
server, but that is only required for Prefect Cloud at this time.  This removes
that restriction, and adds examples of Prefect servers and Prefect Cloud
configurations for reference.


config/samples/v1_prefectworkpool_external_prefect_server.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: PrefectWorkPool
+metadata:
+  labels:
+    app.kubernetes.io/name: prefect-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: external-pool
+spec:
+  type: process
+  server:
+    remoteApiUrl: https://my-server.example.com
```
mitchnielsen, line 11: I noticed that we apply the `/api` suffix if one isn't provided, so I think it's fine to leave it out of the example - but I noticed that Prefect Cloud uses an `api` subdomain too: `api.prefect.cloud/api`. I'm guessing that's just a cloud-specific thing 🤔

mitchnielsen (approved):
Looks good 👍🏼 Picked an example config and got the external work pool set up against Prefect Cloud (dev):

image

config/samples/v1_prefectworkpool_external_prefect_server.yaml
chrisguidry: Yep that's specific to Prefect Cloud, but folks who are hosting it themselves can use any domain they like (but it's always a subpath of `/api`)

api/v1/prefectworkpool_types.go
```diff
@@ -143,11 +143,10 @@ func (s *PrefectWorkPool) Command() []string {
...
 }
 
-// PrefectAPIURL returns the API URL for the Prefect Server.
-// If an API Key is provided, it will return the RemoteAPIURL.
-// Otherwise, it will default to the local, in-cluster API URL.
+// PrefectAPIURL returns the API URL for the Prefect Server, either from the RemoteAPIURL or
+// from the in-cluster server
 func (s *PrefectWorkPool) PrefectAPIURL() string {
-	if s.Spec.Server.APIKey != nil && s.Spec.Server.RemoteAPIURL != nil {
+	if s.Spec.Server.RemoteAPIURL != nil {
```
parkedwards, line 149: ah my fault, i was thinking of Cloud specifically for the remote access scenario

config/samples/v1_prefectworkpool_external_prefect_cloud.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
```
parkedwards, line 1: dumb q, are these `samples/` used anywhere for the e2e tests? or are these just manifests for manual testing?

parkedwards (approved):

config/samples/v1_prefectworkpool_external_prefect_cloud.yaml
chrisguidry: Oh great q I don't know about that, hmmm

mitchnielsen: Yeah don't think they're used anywhere, just there as examples for end users: https://book.kubebuilder.io/getting-started.html?highlight=samples#sample-of-custom-resources

mitchnielsen: They'd probably be helpful in CI when we want to do a full end-to-end test against a real instance, could `kubectl apply` these and confirm the resources come up sucessfully. Would also ensure our examples are functional

PR #71, opened 2024-09-06
Installs pre-commit as part of the toolchain

chrisguidry:


Makefile
```diff
@@ -24,10 +24,19 @@ ENVTEST ?= setup-envtest
...
-tools:
+.PHONY: all
+all: tools build
+
+.PHONY: mise
+mise:
 	@mise install --yes
 
+.git/hooks/pre-commit:
+	@pre-commit install
```
mitchnielsen, line 35: Cool, does this target ensure `pre-commit` gets installed for users who may not have it yet?

.mise.toml
```diff
@@ -8,3 +8,4 @@ kubectl = '1.31'
 kustomize = '5.3.0'
 setup-envtest = '0.17.0'
 yamllint = '1.35.1'
+pre-commit = '3.8.0'
```
mitchnielsen, line 11: Good catch, forgot to add this one 👍🏼

mitchnielsen (approved):

Makefile
chrisguidry: Yep exactly

PR #73, opened 2024-09-13
Create Helm Chart to deploy the `prefect-operator`

jamiezieziula:
This PR generates the initial version of a helm chart to deploy the `prefect-operator`.

### Validation
1. Authenticate to a test cluster
2. In the root directory of this repo, run `helm install prefect-operator ./deploy/charts/prefect-operator`

You should see a `prefect-operator` pod come up healthy after a few minutes. 

### Follow on:
- create release workflow & other relevant maintenance flows
- remove old manifest files & `go` comments for auto-generation of the manifest

Resolves to PLA-306


deploy/charts/prefect-operator/charts/common-2.20.5.tgz
mitchnielsen, line 1: Do we normally commit these to the repo or just let `helm dependency update` resolve them?

deploy/charts/prefect-operator/templates/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
+{{/*
+Copyright Broadcom, Inc. All Rights Reserved.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+
```
mitchnielsen, line 6:
```
```

.mise.toml
```diff
@@ -3,9 +3,11 @@ actionlint = '1.7.1'
...
 golang = '1.21'
 golangci-lint = '1.57.2'
+helm = '3.15'
+helm-ct = '3.11.0'
+helm-docs = '1.13.1'
 kube-controller-tools = '0.14.0'
 kubectl = '1.31'
 kustomize = '5.3.0'
-setup-envtest = '0.17.0'
-yamllint = '1.35.1'
```
mitchnielsen, line 10: I'm guessing yammlint becomes useless when helm enters the picture?

config/default/manager_auth_proxy_patch.yaml
mitchnielsen, line 1: I don't _think_ we can fully delete this yet. There are comments in the Go code ([example](https://github.com/PrefectHQ/prefect-operator/blob/68a8751a0be649a938bc4dbe0348edcb24dc32cd/internal/controller/prefectserver_controller.go#L53-L60)) that the tooling uses to generate these manifests for us.

I'll have to look into how people who build helm charts handle this. Maybe we just ditch the scaffolding helpers entirely and just define them ourselves in Helm. But should probably be a conscious decision, in which case we also remove the related helpers in `Makefile`/etc.

mitchnielsen (approved):
Sweet! I was able to run install the helm chart into a local cluster and all looks good. We may want to add [helm-unittest](https://github.com/helm-unittest/helm-unittest) down the road to make sure all of our configuration works as expected. Looks great as a first iteration though 👍🏼

PR #75, opened 2024-09-16
chore: disable `kube rbac proxy` by default for now

jamiezieziula:
This was part of the original scaffolding we leveraged -- we don't know if we want this yet, so let's disable by default.

Relates to PLA-300


chrisguidry (approved):

mitchnielsen:
I was curious what this was doing for us so I dug in a bit. Turns out this built-in component isn't going to be part of the default scaffolding in the future and so we'll need to pick a migration option. Created https://linear.app/prefect/issue/PLA-318/create-a-plan-for-the-deprecation-of-kube-rbac-proxy.

PR #76, opened 2024-09-17
fix: update workflow to restrict which tag version to pull & set for `prefect-operator` version

jamiezieziula:
Relates to PLA-307

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -28,7 +28,12 @@ jobs:
           # Enable pipefail so git command failures do not result in null versions downstream
           set -x
           echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d%H%M%S')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-release.yaml
```diff
@@ -28,7 +28,12 @@ jobs:
           # Enable pipefail so git command failures do not result in null versions downstream
           set -x
           echo "RELEASE_VERSION=$(date +'%Y.%-m.%-d%H%M%S')" >> $GITHUB_OUTPUT
-          echo "OPERATOR_VERSION=$(git describe --tags `git rev-list --tags --max-count=1`)" >> $GITHUB_OUTPUT
+
+          # This ensures that the latest tag we grab will be of the operator image, and not the helm chart
+          echo "OPERATOR_VERSION=$(\
```
jamiezieziula, line 33: i think this will work -- will publish a release & verify

mitchnielsen (approved):
Looks good to me if it worked in your testing. Kind of gnarly, seems like something the `gh` CLI should help with but I don't think I found anything helpful last time I looked

PR #80, opened 2024-09-19
Remove raw manifest files, update automation to use `helm` instead

jamiezieziula:
Will be good to get @mitchnielsen's opinion on this -- not sure I removed all of the correct `go` comments.


chrisguidry (changes requested):
Wrong repo?  I think we want to retain these for `prefect-operator`, it's how we make the manifests for public release.

PR #82, opened 2024-09-19
Bump golangci-lint version

mitchnielsen:
Bumps golangci-lint to address performance problems that were bringing my machine to a halt.

Related to https://linear.app/prefect/issue/PLA-289/cycle-2-catch-all


chrisguidry (approved):
Thank you!

jamiezieziula (approved):

PR #84, opened 2024-09-19
chore: remove reference to `prefect.yaml`, which was removed in the previous PR

jamiezieziula:
Relates to PLA-309

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -52,7 +52,7 @@ jobs:
         if: github.ref_type == 'tag'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh release upload ${{ github.ref_name }} prefect-crds.yaml prefect.yaml
+        run: gh release upload ${{ github.ref_name }} prefect-crds.yaml
 
   build-and-push-docker-image:
     needs: unit-tests


.github/workflows/release.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
         if: github.ref_type == 'tag'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh release upload ${{ github.ref_name }} prefect-crds.yaml prefect.yaml
+        run: gh release upload ${{ github.ref_name }} prefect-crds.yaml
```
mitchnielsen, line 55: I missed this too - previously, `prefect-crds.yaml` had only the CRDs, and `prefect.yaml` had the CRDS _and_ the Deployment, etc, so folks could add it all in one. We could still populate this by running `helm template` and piping it out to a file, but I'm fine skipping this for now as I bet most will install it via helm.

mitchnielsen (approved):

.github/workflows/release.yaml
jamiezieziula: i'm happy to do that if you think thats useful!

mitchnielsen: can just be a follow-up issue i think, i don't see it as a big priority unless users request it

jamiezieziula: follow up issue [here](https://linear.app/prefect/issue/PLA-334/update-prefect-operator-to-package-manifests-alongside-release)

PR #92, opened 2024-09-23
fix/Fix the output of the image version

jimid27:
Relates to PLA-289

Image version was not properly being saved to the Github output and this should fix. 

Screenshot 2024-09-23 at 10 29 43 AM


chrisguidry (approved):
regex rules everything around me

PR #94, opened 2024-09-24
Manage goimports with mise

mitchnielsen:
Mise allows you to use an experimental feature called "go backend", which will make our tool dependency resolution more consistent.

Related to https://linear.app/prefect/issue/PLA-289/cycle-2-catch-all


chrisguidry (approved):
Amazing, thank you!

PR #97, opened 2024-09-26
Automate Helm Chart release on `operator` release

jamiezieziula:
This PR also:
- updates workflow files to use `_` instead of `-`
- removes `env` specifications where not neccessary
- updates `GITHUB_TOKEN` -> `GH_TOKEN` when using the github cli
- separates `tests` into their own workflow

Closes PLA-349


jimid27 (approved):

mitchnielsen (approved):

jamiezieziula:
FYI - as a follow on - i think we should break out the image build, etc into sep workflow files, similar to how we do this internally for the cloud operator

PR #102, opened 2024-10-01
Allows for adding labels to the generated deployments, services, and jobs

chrisguidry:
This addition gives users fine-grained control over adding extra labels to any
of the deployment, services, or jobs produced by the operator.  This gives folks
control over labels they may need to add, like suppressing Istio sidecars for
specific pods, or adding additional custom labels for application use.


mitchnielsen (approved):
looks great, no notes 👍🏼

PR #105, opened 2024-10-07
feat: configure extra containers for Deployment

mitchnielsen:
Adds a new field to the PrefectServer spec where you can pass a list of extra containers to the Prefect Server Deployment.

Related to https://linear.app/prefect/document/auth-design-doc-7b1ae89e5382

CLOSES CLOUD-422

### DIFF
--- a/api/v1/prefectserver_types.go
+++ b/api/v1/prefectserver_types.go
@@ -39,6 +39,9 @@ type PrefectServerSpec struct {
 	// Resources defines the CPU and memory resources for each replica of the Prefect Server
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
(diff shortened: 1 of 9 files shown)


jakekaplan (approved):
Looks great to my untrained eyes!

chrisguidry (approved):
Looks awesome, what do you think about adding it to `PrefectWorkPool` for completeness?

mitchnielsen:
> what do you think about adding it to PrefectWorkPool for completeness?

Sure thing, I wasn't sure if there was a use case there so happy to add it 👍🏼

mitchnielsen:
Added to WorkPools in ade9ba6.

chrisguidry:
Yeah no use case that I can think of just yet, just want to keep the affordances for both of them the same as much as possible

PR #109, opened 2024-10-09
Remove newline in init function

mitchnielsen:
Removes the newline in the init function. Mostly just making an arbitrary change to trigger a build of the image on `main` that will push the short sha as added in https://github.com/PrefectHQ/prefect-operator/pull/108

Related to https://linear.app/prefect/issue/PLA-343/automatically-update-prefect-operator-helm-chart-versions-in-ccd-to


jakekaplan (approved):

chrisguidry (approved):
That newline was load-bearing!

PR #114, opened 2024-10-11
chore/Optionally add additional ports to the prefect server service

jimid27:
- Relates to: https://linear.app/prefect/issue/PLA-371/cycle-3-catch-all
- Adds optional spec field for additional service ports to include with the prefect server object

### DIFF
--- a/internal/controller/prefectserver_controller_test.go
+++ b/internal/controller/prefectserver_controller_test.go
@@ -550,6 +550,41 @@ var _ = Describe("PrefectServer controller", func() {
 				container := deployment.Spec.Template.Spec.Containers[1]
 				Expect(container.Name).To(Equal("extra-container"))
 			})
+			It("should update the Service with the extra port", func() {
+				// Update the PrefectServer with an extra port
+				Expect(k8sClient.Get(ctx, name, prefectserver)).To(Succeed())
+				prefectserver.Spec.ExtraServicePorts = []corev1.ServicePort{
+					{
+						Name:       "extra-port",
+						Port:       4300,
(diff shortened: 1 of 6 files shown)


internal/controller/prefectserver_controller_test.go
```diff
@@ -550,6 +550,41 @@ var _ = Describe("PrefectServer controller", func() {
 				container := deployment.Spec.Template.Spec.Containers[1]
 				Expect(container.Name).To(Equal("extra-container"))
 			})
+			It("should update the Service with the extra port", func() {
```
mitchnielsen, line 553:
```suggestion

			It("should update the Service with the extra port", func() {
```

internal/controller/prefectserver_controller_test.go
```diff
@@ -550,6 +550,41 @@ var _ = Describe("PrefectServer controller", func() {
...
+				Expect(k8sClient.Get(ctx, name, prefectserver)).To(Succeed())
+				prefectserver.Spec.ExtraServicePorts = []corev1.ServicePort{
+					{
+						Name:       "extra-port",
+						Port:       4300,
+						TargetPort: intstr.FromString("extra-port"),
+						Protocol:   corev1.ProtocolTCP,
+					},
+				}
+				Expect(k8sClient.Update(ctx, prefectserver)).To(Succeed())
```
mitchnielsen, line 564:
```suggestion

				Expect(k8sClient.Update(ctx, prefectserver)).To(Succeed())
```

mitchnielsen (approved):
nice!

PR #121, opened 2024-10-28
Lower migration job TTL

mitchnielsen:
Lowers the job TTL from 7 days to 1 hour.

This should greatly reduce the number of completed Jobs sitting around in the cluster, which will free up a significant number of IPs (and clutter).

We may end up wanting to make this configurable in the PrefectServer spec.

Context: https://linear.app/prefect/document/prefect-load-testing-notes-b87465895445

Related to https://linear.app/prefect/issue/PLA-435/determine-appropriate-ttl-for-jobs-created-by-the-prefect-operator


jimid27 (approved):
🙏

chrisguidry (approved):

PR #126, opened 2024-10-31
Specify CRD changes in source code

mitchnielsen:
Specifies the CRD changes in the source code. A couple recent CRD changes were made directly in the CRD YAML files, but they should be made from the source so that `make manifests` generates the resulting YAML for us.

Related to https://linear.app/prefect/issue/PLA-452/cycle-5-catch-all


prefectcboyd (approved):
nice

parkedwards (approved):

PR #129, opened 2024-11-01
Support setting extra args in Deployment command

mitchnielsen:
Supports setting extra args in the Deployment command.

This will allow us to set `--no-ui` in cases where we don't need to serve the UI.

Related to https://linear.app/prefect/issue/PLA-452/cycle-5-catch-all

### DIFF
--- a/api/v1/prefectserver_types.go
+++ b/api/v1/prefectserver_types.go
@@ -45,6 +45,9 @@ type PrefectServerSpec struct {
 	// ExtraServicePorts defines additional ports to expose on the Prefect Server Service
 	ExtraServicePorts []corev1.ServicePort `json:"extraServicePorts,omitempty"`
 
+	// ExtraArgs defines additional arguments to pass to the Prefect Server Deployment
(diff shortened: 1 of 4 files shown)


prefectcboyd (approved):
nice

api/v1/prefectserver_types.go
```diff
@@ -281,7 +284,10 @@ func (s *PrefectServer) Image() string {
 }
 
 func (s *PrefectServer) Command() []string {
-	return []string{"prefect", "server", "start", "--host", "0.0.0.0"}
+	command := []string{"prefect", "server", "start", "--host", "0.0.0.0"}
+	command = append(command, s.Spec.ExtraArgs...)
```
jimid27, line 288: would this really be `extraCommands` since args are their own field in k8s?

mitchnielsen: hmmm. technically yes, but here we're shoving args into the command field and not specifying any args directly, even though everything after the first one is an arg 🤔

PR #135, opened 2024-11-18
Add make clean

mitchnielsen:
Add Make target to clean up project directory

Adds a Makefile target to clean up the project directory of any
generated files.

The first one added here is to clean up the kubebuilder binaries, which
are write-protected and therefore cannot be removed with 'rm -rf'.

Related to https://linear.app/prefect/issue/PLA-690/cycle-6-catch-all


jamiezieziula (approved):

chrisguidry (approved):
Thanks!

PR #143, opened 2025-01-27
Adds native support for configuring a Redis message broker

chrisguidry:
The Prefect server image now has support for using Redis as a message
broker natively, but it requires setting a few env vars to take
advantage of it.  Here we'll give `prefect-operator` native
configuration for pointing to an existing Redis instance, just like with
PostgreSQL.

This also refactors all of the environment variable generation to one
spot for the `PrefectServer`, keeping things a little easier to follow.


bunchesofdonald (approved):
Nice!

PR #169, opened 2025-06-05
chore(CI): skip docker login on PR workflows

### LINKED ISSUE #156: Pull requests from other contributors can't login to dockerhub
We shouldn't need to log in there just to build images, let's relax that.  See this PR for an example: https://github.com/PrefectHQ/prefect-operator/actions/runs/14684519166

mitchnielsen:
We don't push docker images on PR workflows, so no need to `docker login` either. This was causing problems in fork pipelines because the login credentials weren't available (as expected).

Closes https://github.com/PrefectHQ/prefect-operator/issues/156

image

### DIFF
--- a/.github/workflows/build-and-push-image.yaml
+++ b/.github/workflows/build-and-push-image.yaml
@@ -12,6 +12,11 @@ name: Build and Push Image
         description: Image name (without the tag & registry)
         required: true
         type: string
+      docker-login:
+        description: Whether to log in to Docker Hub
+        required: false
+        type: boolean
+        default: true
       push:
         description: Whether to push the image to the registry
         required: false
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push-image.yaml
```diff
@@ -12,6 +12,11 @@ name: Build and Push Image
         description: Image name (without the tag & registry)
         required: true
         type: string
+      docker-login:
+        description: Whether to log in to Docker Hub
+        required: false
+        type: boolean
+        default: true
```
mitchnielsen, line 19: Default to `true` so we only need to specify `false` when we want to disable it

chrisguidry (approved):
Thanks, dude!

jimid27 (approved):
Ty ty

PR #170, opened 2025-06-09
`PrefectDeployment`

### LINKED ISSUE #164: Prefect Deployment CRD
I think this operator could also manage Prefect Deployments.
The operator would manage deployments by creating, updating and deleting them.
This would also aid managing prefect deployment in GitOps workflows like flux/argo. The resource could implement a similar interface to flux's HelmRelease:

```yaml
apiVersion: prefect.io/v1
kind: PrefectDeployment
metadata:
  labels:
    app.kubernetes.io/name: prefect-operator
    app.kubernetes.io/managed-by: kustomize
  name: myDeployment
status:
  deploymentID: "f31ba293-4028-4492-9219-c4eddf56707e"
  flowID: "e1e69e4b-f767-445e-8cd0-900a3dcbfc2f"
spec:
  interval: 10m # reconcile interval
  deletionPolicy: "delete"
  retryInterval: 5m # retry if reconcile fails
  prefectApiUrl: "http://prefect-server:4200/api"
  deploymentSpec:
    schedule: "0 * * * *"
    entrypoint: "path/to/flow:func"
    storage: None
    tags: 
      - myTag
    jobVariables:
       image: "myimage:latest"
       labels:
         myLabel: myValue
       command: ["prefect.engine"]
       ...
```

What do you think?

chrisguidry:
Adds the new `PrefectDeployment` CRD and a controller for it.  This
CRD will create a Prefect `Deployment` against the given Prefect server,
which requires actually connecting to the running Prefect server for
the first time.

This also adds a Python script to the repository that helps with
generating a `PrefectDeployment` resource from a given flow, analogous
to the `prefect deploy ...` command.  I'm aiming here to reuse as much
from `PrefectHQ/prefect` as possible there.

Note: the CRD spec ends up using `camelCase` names rather than
`snake_case` in order to retain consistency with the other CRDs and with
Kubernetes convetions.

Much of this was generated by Claude (guided by me) in several stages,
and I'll be reviewing this PR as if it were a fresh contribution.

Closes #164

Co-Authored-By: Claude <noreply@anthropic.com>

What Claude itself has to say about the changes:

  🎯 High-Level Summary: PrefectDeployment CRD Implementation

  This PR introduces a major new feature: PrefectDeployment Custom Resource Definition - allowing users to manage Prefect deployments (Python flows)
  declaratively through Kubernetes. While the diff is large (~7k lines), the majority is comprehensive test coverage and generated code.

  📋 What This PR Delivers

  1. New PrefectDeployment CRD (api/v1/prefectdeployment_types.go)
    - Declarative management of Prefect deployments (Python flows)
    - Supports schedules, parameters, work pools, and server connections
    - Full lifecycle management with proper cleanup via finalizers
  2. Production-Ready Controller (internal/controller/prefectdeployment_controller.go)
    - Reconciles K8s resources with Prefect API deployments
    - Handles creation, updates, deletion, and drift detection
    - Robust error handling and status reporting
  3. Prefect API Client (internal/prefect/client.go)
    - HTTP client for Prefect API operations (deployments, flows)
    - Support for both Prefect Cloud and self-hosted servers
    - Port-forwarding for local development scenarios

  🧪 Quality & Testing Focus

  The large line count is primarily comprehensive test coverage:
  - 911 lines: PrefectDeployment controller tests (integration scenarios)
  - 665 lines: Prefect API client tests (HTTP mocking, error cases)
  - 595 lines: Conversion logic tests (100% coverage, edge cases)
  - 466 lines: Server reference tests (API keys, URLs, error handling)

  🏗️ Architecture Improvements

  1. Better Separation of Concerns
    - Moved API key logic from controller to PrefectServerReference domain object
    - Centralized Prefect API conversion logic in dedicated modules
    - Cleaner client creation patterns
  2. Code Simplification
    - Inlined trivial helper functions to reduce indirection
    - More direct method calls instead of verbose wrapper functions
    - Removed unnecessary abstractions
  3. Robust Error Handling
    - Comprehensive error scenarios in tests
    - Proper finalizer-based cleanup to prevent resource leaks
    - Graceful handling of Prefect API failures

  📖 Documentation & Tooling

  - Generated CRD docs (PrefectDeployment.md)
  - Python helper script for generating deployment YAMLs
  - Sample configurations for common use cases

  🎯 Review Focus Areas

  1. Core Logic: prefectdeployment_controller.go (~260 lines) - the main reconciliation logic
  2. API Types: prefectdeployment_types.go (~230 lines) - the CRD structure
  3. Client Interface: internal/prefect/client.go - Prefect API integration
  4. Conversion Logic: internal/prefect/convert.go - K8s ↔ Prefect API mapping

  Everything else is either tests, generated code, or documentation.


api/v1/prefectworkpool_types.go
```diff
@@ -55,36 +54,6 @@ type PrefectWorkPoolSpec struct {
 	DeploymentLabels map[string]string `json:"deploymentLabels,omitempty"`
 }
 
-type PrefectServerReference struct {
```
chrisguidry, line 58: This was hoisted up to be reusable between `PrefectWorkPool` and `PrefectDeployment`

parkedwards (approved):

internal/prefect/client.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+}
+
+// NewClientFromServerReference creates a new PrefectClient from a PrefectServerReference
+func NewClientFromServerReference(serverRef *prefectiov1.PrefectServerReference, apiKey string, log logr.Logger) (*Client, error) {
+	// Create a base client first to check if we're running in cluster
+	baseClient := NewClient("", apiKey, log)
+
+	// Determine if we need port-forwarding
+	needsPortForwarding := !baseClient.isRunningInCluster() && serverRef.IsInCluster()
```
mitchnielsen, line 84: Non-blocking: I don't fully understand the need for this yet, but I'm happy to proceed with it for now. I think I would have suspected that this controller would reach the Prefect API by either the domain on the Ingress (if external) or by the Service (if internal).

mitchnielsen (approved):

internal/prefect/client.go
chrisguidry: Ahh, yes, great question, this is for local development.  When developing the operator, you're running it locally on your laptop against some other cluster (like minikube, etc).  Since the operator isn't actually running inside the cluster where the Prefect server is running, we can't access it with cluster-local addresses, so we need to port-forward to access it.  Totally open to other approaches here, because this does add a lot of code/complexity.

mitchnielsen: Ahh roger that. Maybe in the future we can clarify that somehow but sounds good for now!

chrisguidry: Ah thanks for the tip, I'll check that out!

PR #187, opened 2025-08-29
Delete workpool from server when CRD is deleted

### LINKED ISSUE #184: Deleting work pool CRD does not delete work pool from the server
I created the work pool via CRD and it appeared in the UI. After deleting the CRD I was expecting that the work pool would also be deleted from the server and UI.

Image

Is this by design?

Blarc:
Closes #184


chrisguidry (approved):
Looks great!!  I'll get a release out today

PR #251, opened 2026-02-17
chore: bump Go to 1.26 and apply go fix modernizations

mitchnielsen:
## Summary

Bumps Go to 1.26 and applies the automated fixes from `go fix ./...` to modernize the codebase:

- Use `new()` builtin instead of `k8s.io/utils/ptr.To()` for generic pointer construction (drops `k8s.io/utils` as a direct dependency)
- Replace `interface{}` with the `any` type alias
- Use `maps.Copy` from the standard library instead of manual for-range map copying
- Use `strings.Contains`, `strings.Cut`, and `strings.SplitSeq` instead of older string functions
- Use `for range N` integer range syntax

All changes were produced by `go fix`, just split into separate commits by category.

Related to https://linear.app/prefect/issue/PLA-2340/apply-go-fix-changes-from-go-126

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,8 +1,8 @@
 [tools]
 actionlint = "1.7.10"
 ginkgo = '2.25.2'
-golang = '1.24'
-golangci-lint = "2.8.0"
+golang = '1.26'
+golangci-lint = "2.9.0"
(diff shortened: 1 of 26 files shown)


chrisguidry (approved):
Wonderful!  Thank you!

PR #296, opened 2026-07-08
fix(project): track PrefectDeployment in PROJECT

ShmuelOps:
`PROJECT` (kubebuilder's scaffolding manifest) lists only **PrefectServer** and **PrefectWorkPool** under `resources`, but the repo already ships the **PrefectDeployment** API (`api/v1/prefectdeployment_types.go` + controller + `PrefectDeployment.md`). Add the missing entry so `PROJECT` matches the actual resources.

```yaml
- api:
    crdVersion: v1
    namespaced: true
  controller: true
  domain: prefect.io
  kind: PrefectDeployment
  path: github.com/PrefectHQ/prefect-operator/api/v1
  version: v1
```

Added in the canonical kubebuilder entry format (identical shape to the Server/WorkPool entries). `kubebuilder create api` wasn't used because it would re-scaffold the existing types/controller; this only adds the tracking entry.


mitchnielsen (approved):
Thank you!

PR #294, opened 2026-07-08
feat: add PrefectAutomation CRD + controller

ShmuelOps:
## What

Adds a `PrefectAutomation` CRD (+ controller) so Prefect **automations** can be managed declaratively via the operator — alongside `PrefectServer` / `PrefectWorkPool` / `PrefectDeployment`. The schema covers the full option set of the Terraform provider's `prefect_automation` resource.

## Schema

- `name`, `description`, `enabled`
- **`trigger`** — exactly one of:
  - `event` — `posture` (Reactive/Proactive), `expect`, `after`, `forEach`, `match`, `matchRelated`, `threshold`, `within`
  - `metric` — `metric{name, operator, threshold, range, firingFor}`, `match`, `matchRelated`
  - `compound` — `require` (`any`/`all`/number), `within`, nested `triggers` (event/metric)
  - `sequence` — `within`, nested `triggers` (event/metric)
- **`actions` / `actionsOnTrigger` / `actionsOnResolve`** — all action types, with `source`, `deploymentId`, `deploymentName`, `automationId`, `workPoolId`, `workQueueId`, `scheduleId`, `blockDocumentId`, `parameters`, `jobVariables`, `state`, `name`, `message`, `subject`, `body`, `payload`.

## Deployment references (declarative)

`run-deployment` (and pause/resume-deployment) needs the target deployment's **UUID**, which isn't known declaratively. In addition to `deploymentId` (explicit UUID), an action may set **`deploymentName`** — the controller resolves it to a deployment ID at reconcile time (`POST /deployments/filter`), **requeuing until the named deployment exists**. This lets automations reference deployments by name in GitOps without pre-fetching IDs.

## How

Mirrors the `PrefectDeployment` pattern:
- `api/v1/prefectautomation_types.go` — typed CRD (+ generated deepcopy). Child triggers use a non-recursive `PrefectChildTrigger` (event/metric) to keep the CRD schema structural.
- `internal/prefect/convert_automation.go` — builds the Prefect API trigger/action payloads; resolves `deploymentName` via a name→id map supplied by the controller.
- `internal/prefect/automation.go` — client CRUD (`POST`/`GET`/`PUT`/`DELETE` `/automations`) + `FindDeploymentByName`. Trigger/actions are generic maps so the client stays decoupled from the large Prefect schema.
- `internal/controller/prefectautomation_controller.go` — reconciler with finalizer, spec-hash sync + drift requeue, create-vs-update by stored ID, deployment-name resolution + requeue, status conditions.
[...]

PR #297, opened 2026-07-09
fix: harden PrefectAutomation controller (#294 follow-ups)

mitchnielsen:
## What

Follow-up hardening for the `PrefectAutomation` CRD + controller merged in #294. None of these were blockers, but worth addressing before the CRD sees real use.

- **Self-heal on out-of-band delete** — if the automation is deleted directly in Prefect, `UpdateAutomation` returned a generic error and the controller looped on `SyncError` forever. A 404 on update now surfaces as a sentinel (`ErrAutomationNotFound`); the controller clears the stored ID and recreates on the next pass.
- **Propagate JSON errors** — invalid `parameters` / `jobVariables` JSON in an action was silently dropped, unlike the `match` / `matchRelated` paths. It now returns the error.
- **Enforce deploymentId / deploymentName rules in `Validate()`** — the two are mutually exclusive, and for `run/pause/resume-deployment` the presence of a target is governed by `source`. This mirrors the Prefect API schema: `"selected"` (the default) requires exactly one target, `"inferred"` requires none.
- **Error on ambiguous deployment names** — `FindDeploymentByName` used `limit:1` + first result, so a bare name matching multiple deployments resolved arbitrarily. Deployment names are only unique per flow (canonical reference is `{flow}/{deployment}`), so it now errors on `>1` match instead of picking one silently.
- **Tests** — adds the missing controller envtest suite plus unit coverage for each item above.

## Notes

Full flow-qualified (`flowName/deploymentName`) reference support is a larger CRD change and is left as a follow-up. This PR just makes the ambiguous case fail loudly instead of mis-resolving.

No API-shape changes, so the generated CRD / deepcopy / docs are unchanged.

Related to #294

<details>

Started from a follow-up review of #294 that listed five hardening items. The two low-confidence decisions (how to handle duplicate deployment names, and whether a target is always required for deployment actions) were resolved against the Prefect API schema and docs rather than guessed: deployment names are unique only per flow, and the `source` field defaults to `"selected"` which requires a target while `"inferred"` forbids one.

Verified with `make test` (all suites green, including the new controller specs), `make lint` (clean), and `make manifests generate` (no drift).
</details>

### DIFF
--- a/api/v1/prefectautomation_types.go
+++ b/api/v1/prefectautomation_types.go
@@ -184,6 +184,13 @@ type PrefectSequenceTrigger struct {
 	Triggers []PrefectChildTrigger `json:"triggers"`
 }
 
+// Deployment-target action types (their target is a deployment).
+const (
+	ActionRunDeployment    = "run-deployment"
+	ActionPauseDeployment  = "pause-deployment"
+	ActionResumeDeployment = "resume-deployment"
+)
+
 // PrefectAutomationAction matches Prefect's action union. `type` selects the
 // action; the remaining fields apply to the relevant action types.
 type PrefectAutomationAction struct {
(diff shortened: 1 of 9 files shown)


ShmuelOps (approved):

ShmuelOps:
While testing the self-heal from this PR on a live cluster, I hit a related but separate issue worth flagging: the recreate only happened on the **10-minute drift** (or an operator restart), not promptly.

Root cause is `needsSync`: `PrefectAutomation` persists `Status.LastSyncTime`, so once synced it short-circuits for 10 minutes and doesn't re-check Prefect in between — an out-of-band edit/delete isn't reconciled until the drift window elapses. `PrefectDeployment` reconciles every pass (its status has no `LastSyncTime`), so it self-heals continuously. So this PR's recreate-on-404 works, but only fires at the drift boundary.

I opened #298 to fix the cadence (`needsSync` returns true each reconcile) — orthogonal to this PR (doesn't touch `syncWithPrefect`/`UpdateAutomation`). The two compose: #297 makes delete recreate, #298 makes it (and drift/edit correction) happen within one requeue interval. Happy to fold #298 into this PR instead if you'd prefer them together.

mitchnielsen:
> While testing the self-heal from this PR on a live cluster, I hit a related but separate issue worth flagging: the recreate only happened on the **10-minute drift** (or an operator restart), not promptly.
> 
> Root cause is `needsSync`: `PrefectAutomation` persists `Status.LastSyncTime`, so once synced it short-circuits for 10 minutes and doesn't re-check Prefect in between — an out-of-band edit/delete isn't reconciled until the drift window elapses. `PrefectDeployment` reconciles every pass (its status has no `LastSyncTime`), so it self-heals continuously. So this PR's recreate-on-404 works, but only fires at the drift boundary.
> 
> I opened #298 to fix the cadence (`needsSync` returns true each reconcile) — orthogonal to this PR (doesn't touch `syncWithPrefect`/`UpdateAutomation`). The two compose: #297 makes delete recreate, #298 makes it (and drift/edit correction) happen within one requeue interval. Happy to fold #298 into this PR instead if you'd prefer them together.

Nice, thank you. I'll probably merge this one in first since it's already fairly large and then take a look at yours 👍🏼

jimid27 (approved):

PR #298, opened 2026-07-09
fix: re-sync PrefectAutomation every reconcile for continuous self-heal

ShmuelOps:
## What

`PrefectAutomation` persisted `Status.LastSyncTime`, so `needsSync` short-circuited on a **10-minute drift timer** and never re-checked Prefect in between reconciles. A manual edit or delete of the automation in the Prefect UI/API was not reconciled back until the drift window elapsed or the operator restarted.

`PrefectDeployment` reconciles against Prefect on **every** pass (its status carries no `LastSyncTime`), so it self-heals out-of-band changes continuously. This aligns `PrefectAutomation` with that behavior: `needsSync` now returns `true` each reconcile. `syncWithPrefect` is idempotent (update-in-place), so re-running it every pass is safe.

## Why it was observable

With the drift gate, a deleted automation showed only the `Reconciling` debug line every ~10s and never came back until a restart — even though the CR was `Ready`. After this change the controller re-checks each loop and re-applies desired state.

## Relationship to #297

Complementary. #297 makes the **delete** case recreate (surfaces a 404 on update as a sentinel → recreate). This PR fixes the **cadence** so that recreate (and any drift/edit correction) happens within one requeue interval instead of only on the 10-minute drift. Together they give prompt, continuous self-heal. This PR is orthogonal to #297 (it does not touch `syncWithPrefect`/`UpdateAutomation`) and can merge independently.

## Test

Adds `TestAutomationNeedsSync` — a unit test asserting `needsSync` returns `true` in steady state (id set, spec hash unchanged, generation observed, just synced), which was `false` before.

Related to #294, #297.

### DIFF
--- a/internal/controller/prefectautomation_needssync_test.go
+++ b/internal/controller/prefectautomation_needssync_test.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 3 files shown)


mitchnielsen (commented):
Thanks again!

mitchnielsen (approved):

PR #299, opened 2026-07-09
feat: make Prefect drift-resync interval configurable per resource

mitchnielsen:
## What

Follow-up to #297 / #298. The #298 review surfaced four non-blocking observations about how PrefectAutomation polls Prefect for out-of-band drift. This PR addresses them together by unifying the cadence across the three controllers that reconcile against the Prefect **API** — Automation, Deployment, WorkPool — behind one configurable knob.

Before this, the three behaved differently:

| Controller | Drift cadence before |
|---|---|
| PrefectAutomation | re-synced every ~10s (#298 removed its `LastSyncTime` stamp) |
| PrefectDeployment | re-synced every ~10s (never stamped `LastSyncTime`) |
| PrefectWorkPool | hard-coded 10-minute gate, and didn't requeue on a timer at all |

## How

- **`spec.interval`** (`metav1.Duration`, optional) on `PrefectAutomation` / `PrefectDeployment` / `PrefectWorkPool`, defaulting from a new **`--default-resync-interval`** flag (5m) when unset. The Helm chart exposes the default as `operator.defaultResyncInterval`.
- Each `needsSync` now gates on the effective interval (spec → default → clamped to a 10s floor) instead of a fixed 10m, and requeues with ~10% jitter to avoid a thundering herd. Requeue delays are measured from `LastSyncTime`, so restarts and watch events do not postpone an already-scheduled drift check.
- **`LastSyncTime` is stamped on every successful sync again**, so the field is meaningful (it was permanently null for automation after #298). Following the k8s api-conventions / Flux / cert-manager pattern, the timestamp is written only on a real sync — from the controller, not on every reconcile — so it doesn't churn status.

`PrefectServer` is intentionally untouched: it manages in-cluster workloads, not Prefect API resources, so it has no `needsSync`/drift path.

## Why these choices

I had a subagent research established operator conventions before picking an approach:

- **Configurable interval**: Flux uses a required per-resource `spec.interval`; ArgoCD uses a global `timeout.reconciliation`; the common best-of-both is a spec field with a global-flag default. That's what this implements.
- **`LastSyncTime`**: writing a wall-clock timestamp to status on *every* reconcile is a mild anti-pattern (status-write churn, and this repo has no `GenerationChangedPredicate`). The convention is to write it only on a semantically real event (an actual sync), which is what the controller now does.

## Net effect

Same prompt self-heal as #298 (tunable down to the 10s floor if you want it), without hitting the Prefect API on every ~10s loop, and WorkPool now self-heals on a timer like the others.

## Test

- `internal/utils/resync_test.go` covers floor clamping, spec-vs-default precedence, remaining-delay calculation, and jitter bounds.
- Controller envtest: `spec.interval` overrides the default requeue; `LastSyncTime` is stamped and gates re-sync within the interval; rewrote the #298 `needsSync` unit test around the new interval semantics.
- `make test`, `make lint`, generated manifests/docs, Helm linting, and the Helm unit test are green.

Related to #294, #297, #298

<details>

Grew out of reviewing #298. The four items I raised there (simplify `needsSync`, the vestigial `LastSyncTime`, the WorkPool outlier, and cadence cost) turned out to have a single common cause — a hard-coded interval — so rather than four spot-fixes this unifies them behind a configurable interval, grounded in how Flux/ArgoCD/cert-manager handle the same problem.
</details>

### DIFF
--- a/PrefectAutomation.md
+++ b/PrefectAutomation.md
@@ -144,6 +144,16 @@ Exactly one of event, metric, compound, or sequence must be set.<br/>
           Enabled activates or deactivates the automation<br/>
         </td>
         <td>false</td>
+      </tr><tr>
+        <td><b>interval</b></td>
+        <td>string</td>
+        <td>
+          Interval is how often to re-check this automation against the Prefect API
+to correct out-of-band drift (edits or deletes made directly in Prefect).
+Defaults to the operator's --default-resync-interval when unset. Values
+below 10s are clamped.<br/>
+        </td>
+        <td>false</td>
       </tr></tbody>
 </table>
 
--- a/PrefectDeployment.md
+++ b/PrefectDeployment.md
@@ -104,6 +104,16 @@ PrefectDeploymentSpec defines the desired state of a PrefectDeployment
           WorkPool configuration specifying where the deployment should run<br/>
         </td>
         <td>true</td>
+      </tr><tr>
+        <td><b>interval</b></td>
+        <td>string</td>
+        <td>
+          Interval is how often to re-check this deployment against the Prefect API
+to correct out-of-band drift (edits or deletes made directly in Prefect).
+Defaults to the operator's --default-resync-interval when unset. Values
+below 10s are clamped.<br/>
+        </td>
+        <td>false</td>
       </tr></tbody>
 </table>
 
--- a/PrefectWorkPool.md
+++ b/PrefectWorkPool.md
@@ -112,6 +112,16 @@ PrefectWorkPoolSpec defines the desired state of PrefectWorkPool
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>interval</b></td>
+        <td>string</td>
+        <td>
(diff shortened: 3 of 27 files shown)


ShmuelOps:
Thanks for the great PR!

I love the addition of `spec.interval` on the resources themselves—it aligns perfectly with modern Kubernetes operator patterns (like Flux and External Secrets) for handling drift detection. 

I have one small suggestion: since the operator now accepts a `--default-resync-interval` flag as a fallback, it would be highly beneficial to expose this flag as a dedicated, configurable option in the Helm chart's `values.yaml`. 

While users can technically already pass this in using `extraArgs`, adding an explicit value for it is a relatively small effort that makes tuning the global polling behavior much more discoverable and user-friendly for different environments.

jamiezieziula (approved):

mitchnielsen:
good call! done in [02e23ed](https://github.com/PrefectHQ/prefect-operator/pull/299/commits/02e23ed80fdf620c7afc2c36204ee679056c73ae)

Issue #171, opened 2025-06-11
Adopt the Prefect best practices for an HA server when replicas > 1


chrisguidry:
The documentation (early PR at https://github.com/PrefectHQ/prefect/pull/18279/files) discusses how to enable HA for Prefect servers if you are running PostgreSQL and Redis with a specific combination of API/UI and background services.  Let's make this pattern easier to follow for a `PrefectServer`.

* We should have validation (validating webhook?) when enabling multiple replicas of `PrefectServer` that you're using PostgreSQL and providing a Redis connection
* We can allow users to request an additional background service deployment with its own independent replicas; this and the primary/API/UI deployment could each have the right command line options to run the right subset of services

PR #300, opened 2026-07-13
feat: add server replicas + server/workpool affinity

ShmuelOps:
## Summary
`PrefectServer` is always reconciled with a single replica, and neither `PrefectServer` nor `PrefectWorkPool` exposes pod affinity — so there's no way to run the server/worker with multiple replicas or spread them across nodes for HA. This adds those knobs, bringing the operator to parity with the options the official Helm charts already document.

## Changes
| CRD | New field | Behavior |
|---|---|---|
| `PrefectServer` | `spec.replicas` (`*int32`, default 1) | Sets the Deployment replica count — **Postgres backend only**. sqlite/ephemeral use node-local storage and must stay single-replica; a CEL validation (below) **rejects** `replicas > 1` on those backends instead of silently ignoring it. |
| `PrefectServer` | `spec.affinity` (`corev1.Affinity`) | Wired onto the server pod spec. |
| `PrefectWorkPool` | `spec.affinity` (`corev1.Affinity`) | Wired onto the worker pod spec (replica count already exists as `spec.workers`). |

Use `affinity.podAntiAffinity` to spread replicas across nodes.

### Validation

The `replicas > 1`-requires-Postgres constraint is enforced declaratively on the `PrefectServer` CRD (`x-kubernetes-validations`), so an invalid manifest is rejected at admission:

```yaml
x-kubernetes-validations:
  - rule: '!has(self.replicas) || self.replicas <= 1 || (has(self.postgres) && !has(self.sqlite))'
    message: replicas > 1 is only supported with the postgres backend alone (the controller prefers sqlite when both are set, and sqlite/ephemeral use node-local storage and must run a single replica)
```

### Reconcile-phase check (shared state)

Per review discussion: hard-failing `replicas > 1` without Redis would reject legitimate setups (the damage is feature-conditional), so the controller instead sets a **`SharedStateConfigured` status condition** each reconcile:

| replicas | `spec.redis.leaseStorage` | condition |
|---|---|---|
| 1 | any | `True` / `SingleReplica` |
| >1 | `true` | `True` / `RedisLeaseStorage` |
| >1 | unset/false | `False` / `InMemoryLeaseStorage` (warning: concurrency limits under-enforce) |

Non-blocking — `status.ready` is unaffected.

## Precedent
These mirror knobs the official [prefect-helm](https://github.com/PrefectHQ/prefect-helm) charts already expose:
- prefect-server: `server.replicaCount` (default 1), `server.affinity` (default `{}`)
- prefect-worker: `worker.replicaCount` (default 1), `worker.affinity` (default `{}`)

## Implementation notes
- `Replicas` is applied only in the Postgres deployment-spec path; `Affinity` is applied to all backends' pod specs.
- Regenerated `zz_generated.deepcopy.go` + CRDs (controller-gen). Added a unit test for the `Replicas()` default.
- `make build` / `go vet` / `go test ./api/...` pass.


ShmuelOps:
@mitchnielsen 
Can you review this feature please?

api/v1/prefectserver_types.go
```diff
@@ -30,6 +30,7 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // PrefectServerSpec defines the desired state of a PrefectServer
+// +kubebuilder:validation:XValidation:rule="!has(self.replicas) || self.replicas <= 1 || has(self.postgres)",message="replicas > 1 is only supported with the postgres backend (sqlite/ephemeral use node-local storage and must run a single replica)"
```
mitchnielsen: this will accept `replicas > 1` when `postgres` exists which is good, but the controller happens to look for SQLite configuration first. so a spec that contains both backends + `replicas: 3` would be admitted, but then create a single-replica SQLite deployment

we should probably reject conflicting backends, or require PostgreSQL _without_ SQLite here

what do you think?

api/v1/prefectserver_types.go
mitchnielsen, line 1: I believe we'll need to run `make docs` to generate the `PrefectServer.md` and `PrefectWorkPool.md` docs so they include the new fields you've added

mitchnielsen (changes requested):
sorry for the delay, busy times over here. just a couple small comments

api/v1/prefectserver_types.go
ShmuelOps: Good catch — confirmed: `prefectServerDeployment` dispatches sqlite → postgres → ephemeral, so with both backends set the postgres path (where `Replicas()` is applied) is never reached, and the old rule would admit `replicas: 3` that silently deploys a single-replica sqlite server.

Fixed in 5a0975f by requiring postgres *without* sqlite:

```
!has(self.replicas) || self.replicas <= 1 || (has(self.postgres) && !has(self.sqlite))
```

I went with this narrower rule rather than rejecting conflicting backends outright: a spec with both backends is currently valid (sqlite wins), so rejecting the combination at admission would break existing manifests on CRD upgrade — feels like its own change with a deprecation note. Happy to open that as a follow-up if you'd like.

api/v1/prefectserver_types.go
ShmuelOps: Done — ran `make docs` in 253d090; `PrefectServer.md` / `PrefectWorkPool.md` now include `replicas`, both `affinity` fields, and the (updated) CEL validation.

ShmuelOps:
@mitchnielsen one more design question on `replicas > 1`, from running this in our own cluster: some server-side state defaults to **in-process memory**, so it silently becomes per-replica the moment you scale — most notably concurrency **lease storage** (global/deployment concurrency limits stop being global: each replica hands out leases from its own memory), and similarly the in-memory messaging broker behind events/automations. We had to point the server at `prefect-redis` (`PREFECT_SERVER_CONCURRENCY_LEASE_STORAGE=prefect_redis.lease_storage` + the `PREFECT_REDIS_MESSAGING_*` settings) to make multi-replica behave correctly.

Do you think the operator should treat that as a **blocker** for `replicas > 1` (like the postgres requirement), or is documenting it enough? Enforcing it declaratively is awkward — those knobs arrive via free-form `spec.settings` env entries, so a CEL rule would be name-matching env vars, and users may legitimately supply the settings some other way. My inclination is a note on the `replicas` field docstring ("multi-replica servers also need shared messaging/lease storage, e.g. prefect-redis — in-memory defaults are per-replica") rather than validation, but happy to follow whatever convention you prefer.

mitchnielsen:
Good question. I see what you mean about it being awkward to validate with CEL. But I also don't want to let users get far enough along with an improper setup to the point that the application misbehaves.

What do you think of doing the validation in the controller reconcile loop for now? And then updating the custom resource status to reflect the error. It'll validate later than `kubectl apply`, but at least we'll be giving that feedback before the operator starts configuring resources in a known-bad state.

cc @chrisguidry for any opinions

chrisguidry:
I feel like this PR is walking backwards into the original intent for #171 and I'm very happy this is finally happening!

I'm wondering if, like `sqlite:` vs `postgres:`, we should include a dedicated `redis:` section that, when enabled, configures all of the knobs for redis across the Prefect server.  Then you can use CEL here to validate that in a very similar way to what you're doing with the databases.  I don't think you'd need to do that all in one big PR here, but I think that would resolve the design tension: instead of there being validation on `spec.settings`, it's validation on `spec.redis` being set up.

ShmuelOps:
I'd split it by how unconditional the breakage is:

- **`replicas > 1` requires postgres — keep as plain CEL.** That one is broken for *every* user with no workaround (sqlite/ephemeral are node-local single-writer), so rejecting at apply is correct and can never false-positive.
- **Redis is not in that category.** A multi-replica server without shared lease storage/messaging still works for plain deployments and simple reactive automations — it only degrades when you actually use concurrency limits (leases become per-replica) or proactive/composite triggers (evaluated against one replica's in-memory event stream). Blocking at admission would reject legitimate setups, so the reconcile loop + a status condition feels like the right layer, as @mitchnielsen suggested.

That also composes nicely with @chrisguidry's `spec.redis` follow-up: once Redis is a structured section, the reconcile check becomes a trivial `replicas > 1 && !spec.redis` → status warning instead of name-matching env vars in `spec.settings` — and I'd keep it a warning rather than CEL even then, since the damage is feature-conditional. Happy to wire the status-condition check in a follow-up alongside (or after) `spec.redis`.

chrisguidry:
Nice!  I merged #310 just now, I think you're good to keep rolling here!

ShmuelOps:
@mitchnielsen the reconcile-phase check landed in b4e5bf7, now that #310 gives it a structured field to look at: each reconcile sets a `SharedStateConfigured` status condition — `True`/`SingleReplica`, `True`/`RedisLeaseStorage`, or `False`/`InMemoryLeaseStorage` with a message pointing at `spec.redis.leaseStorage` when `replicas > 1` runs on per-replica in-memory leases. It's a warning condition (doesn't block reconciliation or `status.ready`), per the earlier discussion about the damage being feature-conditional. Unit-tested for all three states.

internal/controller/prefectserver_sharedstate_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	It("is True for a single-replica server", func() {
+		condition := sharedState(&prefectiov1.PrefectServer{})
+		Expect(condition).NotTo(BeNil())
+		Expect(condition.Status).To(Equal(metav1.ConditionTrue))
+		Expect(condition.Reason).To(Equal("SingleReplica"))
+	})
+
+	It("warns for replicas > 1 without redis", func() {
```
mitchnielsen, line 31: I almost wonder if we should error here instead of just warn, but we can start with this for now

mitchnielsen (approved):

PR #302, opened 2026-07-13
fix: manage deployment schedules in place to avoid duplicate scheduled runs

ShmuelOps:
## Problem
Cron-scheduled deployments intermittently produced **duplicate flow runs** for the same minute (the 2nd failing the flow's uniqueness guard). The reconciler re-created the deployment's schedule on essentially every reconcile, minting a **new schedule ID**; Prefect's auto-scheduler keys created runs by `schedule_id`, so when the ID changed between two scheduler passes the same cron tick got scheduled twice.

## Root cause (all confirmed against a live Prefect server)
- `POST /deployments/` (the upsert) with a `schedules` field **replaces** the schedule set → new ID each call; with an **absent** `schedules` field it **wipes** the schedules.
- `PATCH /deployments/{id}` (DeploymentUpdate) **forbids** `name`/`flow_id` (`extra_forbidden` → 422), and returns **204 No Content**.
- A single schedule can be edited only via `PATCH /deployments/{id}/schedules/{id}`, which updates **in place** (preserves the ID, applies the change) — the same call the Prefect UI makes on save.

## Fix
1. Update an existing deployment with `PATCH /deployments/{id}` (POST only for first creation), **without** `schedules` in the body, `name`/`flow_id` stripped, treating a 204/empty response as success.
2. Reconcile schedules separately via the `.../schedules` sub-resource matched by slug: `PATCH` in place / `POST` new / `DELETE` removed. Existing schedules are read via `GetDeployment`.

New client methods + interface + mock; `DeploymentSchedule` gains an `ID` field; adds a `reconcileSchedules` unit test. `go build`/`vet`/`test` pass.

## Validated
Verified end-to-end on a live dev cluster (Prefect 3.6.28): with this change the schedule **ID stays stable across reconciles** (no more duplicate/failed runs) while schedule **edits self-heal in place**. Confirmed both directly against the Prefect API and by running the built controller image; deletion is handled by the existing deployment-level cascade (unchanged).

## Semantics note (PATCH vs the old POST upsert)
Updates now go through `PATCH /deployments/{id}`, which the server applies with `exclude_unset`. So **removing an optional field from the CRD spec no longer clears it** in Prefect (e.g. dropping `paused: true` leaves the deployment paused), whereas the old `POST` upsert reset unset fields to defaults. This is intentional — set fields explicitly to change them; the reconciler no longer clobbers server state for fields the CR doesn't declare.

### DIFF
--- a/internal/controller/prefectdeployment_controller.go
+++ b/internal/controller/prefectdeployment_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -190,6 +191,9 @@ func (r *PrefectDeploymentReconciler) syncWithPrefect(ctx context.Context, deplo
 	var prefectDeployment *prefect.Deployment
 	if deployment.Status.Id != nil && *deployment.Status.Id != "" {
 		prefectDeployment, err = prefectClient.UpdateDeployment(ctx, *deployment.Status.Id, deploymentSpec)
+		if errors.Is(err, prefect.ErrDeploymentNotFound) {
+			prefectDeployment, err = prefectClient.CreateOrUpdateDeployment(ctx, deploymentSpec)
+		}
 	} else {
 		prefectDeployment, err = prefectClient.CreateOrUpdateDeployment(ctx, deploymentSpec)
 	}
@@ -233,46 +237,75 @@ func (r *PrefectDeploymentReconciler) reconcileSchedules(ctx context.Context, cl
 		return fmt.Errorf("get current schedules: %w", err)
 	}
 
-	existingBySlug := make(map[string]prefect.DeploymentSchedule, len(current.Schedules))
-	for _, e := range current.Schedules {
-		if e.Slug != nil {
-			existingBySlug[*e.Slug] = e
-		}
-	}
-
+	desiredBySlug := make(map[string]prefect.DeploymentSchedule, len(desired))
 	var toCreate []prefect.DeploymentSchedule
 	for _, d := range desired {
 		if d.Slug == nil {
 			toCreate = append(toCreate, d)
 			continue
 		}
-		e, ok := existingBySlug[*d.Slug]
-		if !ok {
-			toCreate = append(toCreate, d)
+		desiredBySlug[*d.Slug] = d
+	}
+
+	for _, e := range current.Schedules {
+		d, ok := desiredBySlug[slugValue(e.Slug)]
+		if e.Slug == nil || !ok {
+			if err := client.DeleteDeploymentSchedule(ctx, deploymentID, e.ID); err != nil {
+				return fmt.Errorf("delete schedule: %w", err)
+			}
 			continue
 		}
-		delete(existingBySlug, *d.Slug)
-		if err := client.UpdateDeploymentSchedule(ctx, deploymentID, e.ID, prefect.DeploymentScheduleUpdate{
-			Schedule:         &d.Schedule,
-			Active:           d.Active,
-			MaxScheduledRuns: d.MaxScheduledRuns,
-		}); err != nil {
-			return fmt.Errorf("update schedule: %w", err)
+		delete(desiredBySlug, *e.Slug)
+		if scheduleNeedsUpdate(e, d) {
+			if err := client.UpdateDeploymentSchedule(ctx, deploymentID, e.ID, prefect.DeploymentScheduleUpdate{
+				Schedule:         &d.Schedule,
+				Active:           d.Active,
+				MaxScheduledRuns: d.MaxScheduledRuns,
+			}); err != nil {
+				return fmt.Errorf("update schedule: %w", err)
+			}
 		}
 	}
 
+	for _, d := range desiredBySlug {
+		toCreate = append(toCreate, d)
+	}
 	if len(toCreate) > 0 {
 		if err := client.CreateDeploymentSchedules(ctx, deploymentID, toCreate); err != nil {
 			return fmt.Errorf("create schedules: %w", err)
 		}
 	}
-
-	for _, e := range existingBySlug {
-		if err := client.DeleteDeploymentSchedule(ctx, deploymentID, e.ID); err != nil {
-			return fmt.Errorf("delete schedule: %w", err)
-		}
-	}
 	return nil
+}
+
+func slugValue(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
+}
+
+func scheduleNeedsUpdate(current, desired prefect.DeploymentSchedule) bool {
+	c, d := current.Schedule, desired.Schedule
+	if d.Cron != nil && (c.Cron == nil || *c.Cron != *d.Cron) {
+		return true
+	}
+	if d.RRule != nil && (c.RRule == nil || *c.RRule != *d.RRule) {
+		return true
+	}
+	if d.Interval != nil && (c.Interval == nil || *c.Interval != *d.Interval) {
+		return true
+	}
+	if d.Timezone != nil && (c.Timezone == nil || *c.Timezone != *d.Timezone) {
+		return true
+	}
+	if d.DayOr != nil && (c.DayOr == nil || *c.DayOr != *d.DayOr) {
+		return true
+	}
+	if desired.Active != nil && (current.Active == nil || *current.Active != *desired.Active) {
+		return true
+	}
+	return false
 }
 
 // setCondition sets a condition on the deployment status
--- a/internal/controller/prefectdeployment_controller.go
+++ b/internal/controller/prefectdeployment_controller.go
@@ -184,10 +184,24 @@ func (r *PrefectDeploymentReconciler) syncWithPrefect(ctx context.Context, deplo
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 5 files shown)


internal/controller/prefectdeployment_controller.go
```diff
@@ -184,13 +184,27 @@ func (r *PrefectDeploymentReconciler) syncWithPrefect(ctx context.Context, deplo
 		return ctrl.Result{}, err
 	}
 
-	prefectDeployment, err := prefectClient.CreateOrUpdateDeployment(ctx, deploymentSpec)
+	desiredSchedules := deploymentSpec.Schedules
+	deploymentSpec.Schedules = nil
+
+	var prefectDeployment *prefect.Deployment
+	if deployment.Status.Id != nil && *deployment.Status.Id != "" {
```
chrisguidry, line 192: Once `Status.Id` is set we only ever PATCH, so if someone deletes the deployment in the Prefect UI, every sync 404s from then on and the resource never recovers. The old POST upsert healed this, and the 10-minute drift resync exists for exactly this case. Can we fall back to `CreateOrUpdateDeployment` when the PATCH returns 404?

internal/controller/prefectdeployment_controller.go
```diff
@@ -213,6 +227,54 @@ func (r *PrefectDeploymentReconciler) syncWithPrefect(ctx context.Context, deplo
...
+		}
+	}
+
+	if len(toCreate) > 0 {
+		if err := client.CreateDeploymentSchedules(ctx, deploymentID, toCreate); err != nil {
+			return fmt.Errorf("create schedules: %w", err)
+		}
+	}
+
+	for _, e := range existingBySlug {
```
chrisguidry: This only deletes leftovers that have slugs, so a schedule added by hand in the UI (where slug is optional) lingers forever — while a slugged manual addition gets deleted. The old POST replaced the whole set. Since the CRD requires a slug on every schedule we manage, anything unslugged isn't ours and should go too: iterate `current.Schedules` for the deletes instead of the leftover map.

internal/controller/prefectdeployment_controller.go
```diff
@@ -213,6 +227,54 @@ func (r *PrefectDeploymentReconciler) syncWithPrefect(ctx context.Context, deplo
...
+			toCreate = append(toCreate, d)
+			continue
+		}
+		e, ok := existingBySlug[*d.Slug]
+		if !ok {
+			toCreate = append(toCreate, d)
+			continue
+		}
+		delete(existingBySlug, *d.Slug)
+		if err := client.UpdateDeploymentSchedule(ctx, deploymentID, e.ID, prefect.DeploymentScheduleUpdate{
```
chrisguidry: Not blocking: the server deletes all future auto-scheduled runs on every schedule PATCH, even a no-op ([api/deployments.py#L1324-L1330](https://github.com/PrefectHQ/prefect/blob/0e7435055e18952aa8604dab78507b087a18defb/src/prefect/server/api/deployments.py#L1324-L1330)). We already have the current schedule from the GET — compare it to the desired one and skip the PATCH when nothing changed. That ends the remaining scheduled-run churn instead of just making it idempotent.

internal/prefect/client.go
```diff
@@ -436,6 +449,10 @@ func (c *Client) UpdateDeployment(ctx context.Context, id string, deployment *De
 		return nil, fmt.Errorf("API request failed with status %d: %s", resp.StatusCode, string(body))
 	}
 
+	if resp.StatusCode == http.StatusNoContent || len(bytes.TrimSpace(body)) == 0 {
+		return &Deployment{ID: id, FlowID: deployment.FlowID}, nil
```
chrisguidry: This stamps `FlowID` from the spec, but we just stripped `flow_id` out of the PATCH body — the server never applied it. When the resolved flow differs from what's in Prefect (add/remove of `flowName`, renaming the function part of the entrypoint), `Status.FlowId` claims a linkage that doesn't exist. Either GET the deployment after a 204, or compare the resolved flow ID against `Status.FlowId` upstream and fail the sync condition instead.

internal/prefect/mock.go
```diff
@@ -326,6 +326,76 @@ func (m *MockClient) DeleteDeployment(ctx context.Context, deploymentID string)
 	return nil
 }
 
+func (m *MockClient) CreateDeploymentSchedules(ctx context.Context, deploymentID string, schedules []DeploymentSchedule) error {
```
chrisguidry, line 329: The real server 409s on duplicate slugs within a deployment; the mock appends without checking, so a controller bug that double-creates would pass tests. Worth adding the uniqueness check.

internal/controller/schedule_reconcile_test.go
```diff
@@ -0,0 +1,74 @@
...
+package controller
+
+import (
+	"context"
+	"testing"
+
+	"github.com/PrefectHQ/prefect-operator/internal/prefect"
+)
+
+func TestReconcileSchedules(t *testing.T) {
```
chrisguidry: Everything else in this repo is Ginkgo/Gomega — could this move into the controller suite as specs? The license header is also truncated relative to the other files.

chrisguidry (changes requested):
I checked the root cause against the server code: the scheduler keys runs by `scheduled {deployment_id} {schedule_id} {date}` ([models/deployments.py#L962](https://github.com/PrefectHQ/prefect/blob/0e7435055e18952aa8604dab78507b087a18defb/src/prefect/server/models/deployments.py#L962)), so keeping schedule IDs stable is the right fix. Going through the `/schedules` sub-resource client-side is also more portable than relying on the server-side slug matching in `PATCH /deployments/{id}`, which older 3.x servers don't have.

Focusing only on what this PR changes (there are pre-existing issues around sync frequency I'll file separately): one thing to fix before merge — the 404 case — and a few smaller ones inline.

One semantics change worth a line in the PR description: the server applies `DeploymentUpdate` with `exclude_unset` ([models/deployments.py#L311-L316](https://github.com/PrefectHQ/prefect/blob/0e7435055e18952aa8604dab78507b087a18defb/src/prefect/server/models/deployments.py#L311-L316)), so removing an optional field from the CRD spec no longer clears it in Prefect. Drop `paused: true` from the spec and the deployment stays paused, where the old POST upsert reset it. That may be fine, but it should be deliberate rather than a side effect.

chrisguidry:
Also, thank you for the contributions!!

internal/controller/prefectdeployment_controller.go
ShmuelOps: Done — on a 404 `UpdateDeployment` now returns `ErrDeploymentNotFound` and the reconciler falls back to `CreateOrUpdateDeployment`, so a deployment deleted in the UI is recreated instead of 404ing forever.

internal/controller/prefectdeployment_controller.go
ShmuelOps: Done — the delete loop now iterates `current.Schedules` and removes anything not in the desired set, including slug-less manual additions.

internal/controller/prefectdeployment_controller.go
ShmuelOps: Done — added `scheduleNeedsUpdate`; we compare the current schedule to the desired one (fields the CR sets) and skip the PATCH when nothing changed, so we no longer drop future auto-scheduled runs on every reconcile.

internal/prefect/client.go
ShmuelOps: Done — on a 204 we now `GET` the deployment and return its real `flow_id` instead of stamping it from the (stripped) request body.

internal/prefect/mock.go
ShmuelOps: Done — the mock now returns a 409-style error on a duplicate slug within a deployment.

internal/controller/schedule_reconcile_test.go
ShmuelOps: Done — rewritten as Ginkgo specs in the controller suite, with the full license header. Added a spec covering the unslugged-prune case.

ShmuelOps:
Thanks for the thorough review @chrisguidry — all addressed in the latest commit:

- **404** → `UpdateDeployment` returns `ErrDeploymentNotFound`; the reconciler falls back to `CreateOrUpdateDeployment`, so a UI-deleted deployment is recreated.
- **Unslugged leftovers** → delete loop iterates `current.Schedules` and prunes anything not desired, slug or not.
- **No-op PATCH churn** → `scheduleNeedsUpdate` skips the PATCH when nothing the CR sets changed (no longer drops future runs each reconcile).
- **Fake `flow_id` on 204** → now `GET`s the deployment for the real `flow_id`.
- **Mock** → 409 on duplicate slug.
- **Tests** → moved to Ginkgo specs (+ full license header, unslugged-prune case).
- Added the `exclude_unset` semantics note to the description.

ShmuelOps:
On the `exclude_unset` point: yes, that's deliberate. Going through `PATCH`, the operator only asserts fields the CR actually declares and leaves everything else (like a UI-set `paused`) untouched — I'd rather not clobber server state the CR doesn't own. Documented it under a **Semantics** note in the description. And thanks again — happy to keep iterating!

chrisguidry (approved):

PR #306, opened 2026-07-13
ci: speed up multi-platform image builds

mitchnielsen:
The ARM build spent roughly 17 minutes compiling through QEMU. Use Go's native cross-compilation, reuse BuildKit layers, and cancel superseded PR runs while retaining amd64 and arm64 image validation.

```text
Before: native runner -> QEMU -> ARM Go compiler
After:  native runner -> Go cross-compiler -> ARM binary
```

<details>

A recent PR image build spent 1,027 seconds in the arm64 compile step, compared with 112 seconds for amd64. A clean local multi-platform build completed in roughly 35 seconds after this change, and both resulting binary architectures were verified.
</details>

### DIFF
--- a/.github/workflows/build-and-push-image.yaml
+++ b/.github/workflows/build-and-push-image.yaml
@@ -34,9 +34,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v7
 
-      - name: Set up QEMU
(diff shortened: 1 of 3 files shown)


chrisguidry (approved):
Outstanding

PR #311, opened 2026-07-26
feat: add PrefectWorkQueue resource and controller

ShmuelOps:
## Summary

`PrefectWorkQueue` is a new full-fledged resource for managing work queues declaratively (reworked from the original spec.workQueues-on-PrefectWorkPool approach per review feedback).

A queue named by a `PrefectDeployment`'s `workQueue` field is created implicitly by Prefect with **no concurrency limit**, and there was no CRD surface to set one — so the only declarative way to bound concurrency for those runs was a **deployment-level** limit. Those two are not interchangeable whenever a backlog forms:

- **Work-queue limit** — workers pull runs via a query sorted `NEXT_SCHEDULED_START_TIME_ASC` (`server/models/work_queues.py`), so a backlog drains **oldest-scheduled first**.
- **Deployment-level limit** — when no slot is free the transition is rejected and the run is re-scheduled as `AwaitingConcurrencySlot` with `scheduled_time=now() + concurrency_slot_wait_seconds` (`server/orchestration/core_policy.py`). That **overwrites the ordering key**, so waiting runs drain in whatever order their retries happen to land.

For any pipeline whose batches must be applied in chronological order, only the first behaviour is safe.

## What this adds

```yaml
apiVersion: prefect.io/v1
kind: PrefectWorkQueue
spec:
  server:
    namespace: prefect
  name: ingest
  workPoolName: example-pool
  concurrencyLimit: 1
```

Spec fields mirror the Terraform provider's `prefect_work_queue` resource: `name`, `workPoolName` (immutable via CEL — a queue cannot move between pools), `concurrencyLimit`, `priority`, `description`, `isPaused`, plus the operator-standard `server` reference and per-resource drift `interval`.

The controller follows the `PrefectAutomation` pattern: finalizer-based remote cleanup, spec-hash change detection, interval-based drift resync, and recreate-when-deleted-out-of-band. Two deliberate behaviours:

- **Adoption**: if the queue already exists (e.g. implicitly created by a deployment), the controller adopts it by name rather than failing, then manages it by ID (`GET`/`PATCH`/`DELETE /work_queues/{id}`, so renames work).
- **Declared-fields-only**: fields left unset in the spec are neither sent nor compared, so declaring only a concurrency limit never fights a description or paused flag managed elsewhere.

The client gains `GetWorkQueue` / `GetWorkQueueByID` / `CreateWorkQueue` / `UpdateWorkQueue` / `DeleteWorkQueue`, following the existing method shapes (getters return `(nil, nil)` on 404 like `GetWorkPool`; update returns `ErrWorkQueueNotFound` on 404 like `UpdateAutomation`).

## Testing

- Nine envtest specs: create, adopt-and-patch-only-drifting-fields, no-PATCH-on-matching-resync, spec-change-keeps-ID, recreate-after-out-of-band-delete, sync-error handling, finalizer deletion, workPoolName immutability, ignore-removed. Full suite passes; `golangci-lint` clean; CRDs, deepcopy and `make docs` regenerated.
- Validated end-to-end against a real Prefect 3.6 server and EKS cluster: create, adopt (undeclared description preserved, declared limit applied), out-of-band edit corrected on resync, out-of-band delete recreated under a new ID, CEL rejection of a pool change, and CR deletion removing the remote queue.


chrisguidry:
Thanks for another awesome submission!  Managing work queues on a pool is a great idea, but I'm thinking that we should make these a new full-fledged `PrefectWorkQueue` resource.  I think this would help reconcile some issues with both `PrefectDeployment` and `PrefectAutomation` in how they reference queues.  What do you think?  Would you be up for tackling that?

ShmuelOps:
@chrisguidry Done — reworked as a full-fledged `PrefectWorkQueue` resource with its own controller (following the `PrefectAutomation` pattern: finalizer cleanup, spec-hash change detection, drift resync). Spec fields mirror the Terraform provider's `prefect_work_queue`. Two behaviours worth calling out for review: the controller *adopts* a queue that already exists (the implicitly-created-by-a-deployment case) instead of failing, and it only compares/sends declared fields, so a CR that sets just a concurrency limit won't fight anything else managing the queue. `workPoolName` is immutable via CEL since queues can't move between pools. PR description updated; happy to iterate on how `PrefectDeployment`/`PrefectAutomation` should reference these in a follow-up.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,207 @@
...
+	"net/http"
+)
+
+// ErrWorkQueueNotFound is returned when a work queue no longer exists in
+// Prefect (e.g. it was deleted out-of-band). Callers can use errors.Is to
+// distinguish this from other API errors and recreate rather than loop.
+var ErrWorkQueueNotFound = errors.New("work queue not found")
+
+// WorkQueueSpec is the create/update payload for a work queue.
+type WorkQueueSpec struct {
```
chrisguidry, line 65: Two notes on the JSON tags.

The pointer plus `omitempty` combination is correct for the declared-fields-only design, and it depends on server behavior worth recording here. Both Prefect update paths call `model_dump_for_orm(exclude_unset=True)` (`models/work_queues.py:382`, `models/workers.py:789`), so an omitted field is left alone. `WorkQueueUpdate.is_paused` is a plain `bool` defaulting to `False` (`actions.py:989-999`), not `Optional[bool]`, so if the server used a plain `model_dump()` every PATCH this client sends would unpause the queue. Add a comment saying the PATCH semantics depend on `exclude_unset`, so nobody switches these to value types.

The gap is that `omitempty` can express "unset" and "a value" but not "explicitly null". Because `exclude_unset` keeps explicit nulls, `"concurrency_limit": null` is a supported way to clear a limit in Prefect, and this client cannot send it. See the note on the CRD field.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,207 @@
...
+	ConcurrencyLimit *int32  `json:"concurrency_limit,omitempty"`
+	Priority         *int32  `json:"priority,omitempty"`
+	WorkPoolName     string  `json:"work_pool_name,omitempty"`
+}
+
+// GetWorkQueue retrieves a work queue within a pool by name via
+// GET /work_pools/{pool}/queues/{name}. Returns (nil, nil) when the queue does
+// not exist, matching GetWorkPool's contract.
+func (c *Client) GetWorkQueue(ctx context.Context, workPoolName, name string) (*WorkQueue, error) {
+	url := fmt.Sprintf("%s/work_pools/%s/queues/%s", c.BaseURL, workPoolName, name)
```
chrisguidry: The pool and queue names go into the path unescaped, and here they land in two separate segments.

`types/names.py:74` sets `BANNED_CHARACTERS = ["/", "%", "&", ">", "<"]`. Spaces, `#`, and `?` are all legal in a Prefect queue name. A queue named `batch #2` builds `GET .../queues/batch #2`, and Go treats everything from the `#` as a fragment and never sends it. `?` turns the rest into a query string.

Wrap both segments in `url.PathEscape`. Same for `CreateWorkQueue` at line 104. The existing work pool methods in `client.go` have the same gap, so this is not a regression, but this is the first endpoint where a user-chosen name reaches two path segments. Worth a follow-up issue for `client.go:680`, `:765`, and `:808`.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	c.log.V(1).Info("Work queue created successfully", "workQueueId", result.ID)
+	return &result, nil
+}
+
+// UpdateWorkQueue updates a work queue via PATCH /work_queues/{id}. Fields left
+// unset in the spec are omitted from the payload, so PATCH semantics leave them
+// unchanged in Prefect.
+func (c *Client) UpdateWorkQueue(ctx context.Context, id string, queue *WorkQueueSpec) error {
+	url := fmt.Sprintf("%s/work_queues/%s", c.BaseURL, id)
```
chrisguidry: This is the blocker. `CreateWorkQueue` uses the pool-scoped route and this uses the legacy `/work_queues/{id}` route. They are backed by different model functions and they do not behave the same.

`PATCH /work_pools/{pool}/queues/{name}` calls `bulk_update_work_queue_priorities` whenever `priority` is in the update (`models/workers.py:872-877`), which re-sequences the pool so priorities stay unique. `PATCH /work_queues/{id}` writes the value straight to the row with a plain `sa.update(...).values(**update_data)` (`models/work_queues.py:420-424`) and never renormalizes. So this method can put two queues in one pool at priority 1, which Prefect's own API has no way to produce. Workers poll queues in priority order, so the tie then breaks arbitrarily.

Note the asymmetry: `CreateWorkQueue` with an explicit priority does renormalize (`models/workers.py:565-569`). So `priority: 2` set at create time is clean and the same value set later on an adopted queue is not.

I would switch this and `DeleteWorkQueue` to the pool-scoped routes. `workPoolName` is immutable and the queue name is in the spec, so the only thing the ID buys is renames, and rename is untested (see the test file comment). If you keep the ID routes, `spec.priority` needs a warning in the CRD docs saying the operator writes it without reordering the pool.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,207 @@
...
+		return ErrWorkQueueNotFound
+	}
+	if !isSuccessStatusCode(resp.StatusCode) {
+		return fmt.Errorf("API request failed with status %d: %s", resp.StatusCode, string(body))
+	}
+	return nil
+}
+
+// DeleteWorkQueue deletes a work queue via DELETE /work_queues/{id}.
+func (c *Client) DeleteWorkQueue(ctx context.Context, id string) error {
```
chrisguidry: Same route split, with a second consequence.

`DELETE /work_pools/{pool}/queues/{name}` goes through `models/workers.py:895-928`, which catches the foreign key error and raises `Can't delete a pool's default queue.`, then renormalizes the remaining priorities. `DELETE /work_queues/{id}` is a raw `session.execute(delete(...))` (`models/work_queues.py:484-491`) with neither.

So deleting a CR that adopted a pool's `default` queue trips an unhandled IntegrityError on the foreign key from `work_pool.default_queue_id`, which surfaces as a server error rather than the clean rejection. `handleDeletion` logs it and removes the finalizer anyway, so the CR vanishes, the operator log gets a stack trace, and the queue stays in Prefect. Every other delete leaves a gap in the pool's priority sequence.

chrisguidry (commented):
Reviewed the diff and cross-referenced it against `prefect` at `3.8.1.dev3`. I did not run the test suite.

The shape is right. The controller mirrors `prefectautomation_controller.go` closely: finalizer, spec-hash gate, `needsSync` with the same four checks in the same order, `LastSyncTime` drift resync, `JitterResyncInterval` on success, `RequeueIntervalError` on failure, non-blocking remote delete. Adopt-an-existing-remote-by-name is what `PrefectWorkPool` already does. A top-level CRD rather than `spec.workQueues` on `PrefectWorkPool` is the right call.

The motivation in the description holds up against the source:

- `query_components.py:195-197`: `get_scheduled_flow_runs_from_work_queues` ends in `.order_by(next_scheduled_start_time, id)`, so a queue limit drains oldest-scheduled first.
- `core_policy.py:667-673`: the deployment limit calls `reject_transition` with `AwaitingConcurrencySlot(scheduled_time=now("UTC") + concurrency_slot_wait_seconds)`, which rewrites the sort key. There are no work queue rules in `core_policy.py`.
- `api/deployments.py:135-144`: a deployment with `work_pool_name` and `work_queue_name` creates the queue through `_get_work_queue_id_from_name(create_queue_if_not_found=True)`, which builds `WorkQueueCreate(name=...)` with no concurrency limit and no priority.

One blocker: the client mixes two Prefect route families that do not behave the same. Details inline on `UpdateWorkQueue` and `DeleteWorkQueue`.

`internal/.DS_Store` is committed. Remove it, and add `.DS_Store` to `.gitignore` in the same commit.

`internal/prefect/workqueue.go` has no unit tests. That is 207 lines at zero coverage, and `automation.go` has `automation_test.go` covering the same shape. The 404 branches drive recreate and cleanup: `getWorkQueue` returning `(nil, nil)`, `UpdateWorkQueue` returning `ErrWorkQueueNotFound`, and `DeleteWorkQueue` treating 404 as success. None of the real implementations run today.

Three files outside the diff need updating to match the other kinds:

- `PROJECT` has an entry per kind and is missing `PrefectWorkQueue`.
- `README.md` has a section per CRD linking its generated doc. Add one for `PrefectWorkQueue.md`.
- `deploy/samples/` has a sample per kind. The YAML in the description is a good starting point.

api/v1/prefectworkqueue_types.go
```diff
@@ -0,0 +1,126 @@
...
+	// WorkPoolName is the work pool this queue belongs to. A queue cannot move
+	// between pools, so this field is immutable.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="workPoolName is immutable"
+	WorkPoolName string `json:"workPoolName"`
+
+	// ConcurrencyLimit caps how many flow runs this queue may have running at
+	// once. Unset leaves the queue unlimited.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	ConcurrencyLimit *int32 `json:"concurrencyLimit,omitempty"`
```
chrisguidry, line 63: `Unset leaves the queue unlimited` is true only on create. After the operator has applied a limit, removing the field from the CR does nothing: the spec hash changes, `syncWithPrefect` runs, `workQueueMatches` skips the unset field, no PATCH goes out, and the resource reports `Ready: true` with the old limit still in force.

Prefect supports the clear (`"concurrency_limit": null` survives `exclude_unset`), so the limitation is ours. The fix is to record the last-applied field set in status and send an explicit null for fields that go away. If that is out of scope here, state it on the field and open an issue.

Same applies to `description` and `isPaused`.

api/v1/prefectworkqueue_types.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Priority of this queue within the pool; lower numbers are served first.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	Priority *int32 `json:"priority,omitempty"`
```
chrisguidry, line 75: Worth documenting what setting this does, given the route problem in `UpdateWorkQueue`. Priority is pool-wide state, not per-queue state: Prefect keeps it unique and sequential across the pool, and every change reshuffles the other queues. Two `PrefectWorkQueue` CRs in the same pool declaring the same priority is a configuration nothing here rejects.

api/v1/prefectworkqueue_types.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Conditions store the status conditions of the PrefectWorkQueue instances
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path="prefectworkqueues",singular="prefectworkqueue",shortName="pwq",scope="Namespaced"
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Work Pool",type="string",JSONPath=".spec.workPoolName",description="The work pool this queue belongs to"
```
chrisguidry, line 137: Shows pool, ready, and ID. `spec.name` and `spec.concurrencyLimit` are the two things you want in `kubectl get pwq`, since `metadata.name` and the queue name can differ.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,311 @@
...
+			workQueue.Status.Ready = false
+			if updateErr := r.Status().Update(ctx, workQueue); updateErr != nil {
+				log.Error(updateErr, "Failed to update work queue status", "workQueue", workQueue.Name)
+			}
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+	} else {
+		// Look the queue up by name first: a queue referenced by a deployment is
+		// created implicitly by Prefect, and we adopt it rather than fail.
+		remote, err = prefectClient.GetWorkQueue(ctx, workQueue.Spec.WorkPoolName, workQueue.Spec.Name)
```
chrisguidry: Worth a distinct condition for the missing work pool case, matching how the automation controller handles a missing deployment (`DeploymentNotFound` plus a 5s requeue).

Today, if the `PrefectWorkPool` has not reconciled yet, `GET /work_pools/{pool}/queues/{name}` 404s because the pool is missing, which is indistinguishable from the queue being missing (`api/workers.py:98-102`). The code falls through to `CreateWorkQueue`, which 404s with `Work pool "X" not found.`, and that lands in the condition as a raw `SyncError` on the error interval. The message is legible, but `WorkPoolNotFound` would make the ordering dependency explicit.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,311 @@
...
+		log.Error(err, "Failed to sync work queue with Prefect", "workQueue", workQueue.Name)
+		r.setCondition(workQueue, PrefectWorkQueueConditionSynced, metav1.ConditionFalse, "SyncError", err.Error())
+		workQueue.Status.Ready = false
+		if updateErr := r.Status().Update(ctx, workQueue); updateErr != nil {
+			log.Error(updateErr, "Failed to update work queue status", "workQueue", workQueue.Name)
+		}
+		return ctrl.Result{RequeueAfter: RequeueIntervalError}, nil
+	}
+
+	workQueue.Status.Id = &remote.ID
```
chrisguidry, line 232: `workQueue.Status.Id = &remote.ID` is safe today because every path that leaves `err == nil` also leaves `remote` non-nil, but that is four branches up. A guard, or a comment saying why it holds, would protect this against a future client change.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,311 @@
...
+
+	log.Info("Successfully synced work queue with Prefect", "workQueueId", remote.ID)
+	return ctrl.Result{RequeueAfter: utils.JitterResyncInterval(r.resyncInterval(workQueue))}, nil
+}
+
+// workQueueMatches reports whether the remote queue already has the declared
+// values. Fields left unset in the spec are not compared (and not sent), so
+// declaring only a concurrency limit never fights whatever else is set on the
+// queue.
+func workQueueMatches(desired prefectiov1.PrefectWorkQueueSpec, remote *prefect.WorkQueue) bool {
```
chrisguidry, line 287: The comment explains the intent well. Add the consequence: because unset fields are neither compared nor sent, removing a field from the spec is a no-op that still reports `Ready`.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,311 @@
...
+		if prefectClient == nil {
+			var err error
+			prefectClient, err = prefect.NewClientFromK8s(ctx, &workQueue.Spec.Server, r.Client, workQueue.Namespace, log)
+			if err != nil {
+				log.Error(err, "Failed to create Prefect client for deletion", "workQueue", workQueue.Name)
+				prefectClient = nil
+			}
+		}
+		if prefectClient != nil {
+			if err := prefectClient.DeleteWorkQueue(ctx, *workQueue.Status.Id); err != nil {
```
chrisguidry: This deletes the remote queue whether the operator created it or adopted it, and adoption is the common case. Someone who declares a CR only to put a limit on a queue their deployment already created, then deletes the CR, loses the queue and everything routing through it.

`PrefectWorkPool` does the same thing, so this is consistent. The difference is that pools are declared explicitly and queues get created by deployments. Consider recording in status whether the queue was created or adopted, and skipping the remote delete for adopted ones.

internal/controller/prefectworkqueue_controller_test.go
```diff
@@ -0,0 +1,316 @@
...
+			Expect(err).To(HaveOccurred())
+
+			By("The queue is gone from Prefect")
+			remote, err := mockClient.GetWorkQueueByID(ctx, queueID)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(remote).To(BeNil())
+		})
+	})
+
+	Context("When workPoolName is changed", func() {
```
chrisguidry, line 503: Good set. The `UpdateWorkQueueCalls` assertion for the matching-resync case is the right check, and the adopt test proving the undeclared description survives covers the case that matters.

Missing: a rename. The type comment and the description both say renames work because the queue is managed by ID, but nothing changes `spec.name`. The mock's `UpdateWorkQueue` sets `Name` unconditionally, so a test would prove the claim end to end. Worth deciding what should happen when the rename target already exists in the pool, since that returns a 409 and will loop on `SyncError` with no distinct condition.

ShmuelOps:
Thanks for the thorough review — everything is addressed in the two commits now on the branch. Point by point:

**The blocker (route families).** `UpdateWorkQueue` and `DeleteWorkQueue` now use the pool-scoped routes, per your suggestion. The queue is keyed by `(workPoolName, name)` end to end: the PATCH payload never carries `name`, so nothing can rename. Changing `spec.name` instead stops managing the old queue (left untouched — never deleted, since undeclared queues are never touched) and creates-or-adopts under the new name; that's documented on the field and covered by a test that asserts the old queue survives unmodified. The rename-conflict 409 you asked about can no longer happen: an existing queue under the new name is adopted, not re-created. One live observation: on Prefect 3.6.x the pool-scoped delete of a pool's default queue still surfaces as a 500 rather than the clean `Can't delete a pool's default queue` you saw at 3.8.1.dev3 — the delete stays non-blocking either way, and a default queue would virtually always be adopted (see below), so it's never deleted by the operator anyway.

**Path escaping.** Every user-chosen segment goes through `url.PathEscape` (`workQueueURL`), with a test covering a name with a space, `#`, and `?`. Filed #313 for the same gap in the pool methods in `client.go`.

**Clearing removed fields.** Implemented the way you suggested: `status.appliedFields` records what the last sync declared, and a field that goes away is sent back at its create-time default — explicit `null` for `concurrencyLimit` and `description`, explicit `false` for `isPaused` (it's a plain `bool` server-side, so `null` would 422). `priority` is the exception and is documented as such: it has no create-time default to restore (Prefect keeps pool priorities unique and sequential), so removing it keeps the last value. The field docs on the CRD spell out each behavior, and `WorkQueueSpec` now carries your suggested comment that the pointer-`omitempty` shape depends on the server's `exclude_unset` semantics (with the `is_paused` unpause hazard called out). We'd also hit exactly this bug with `PrefectDeployment.spec.deployment.concurrencyLimit` in production use, so #314 applies the same `appliedFields` + explicit-null mechanism there.

**Deletion of adopted queues.** Agreed the pool-consistency argument didn't hold for queues. `status.adopted` records whether the queue pre-existed at first reconcile; deleting a CR that adopted its queue leaves the queue in Prefect (logged), and only operator-created queues are deleted remotely. Both paths are tested.

**Missing pool.** `CreateWorkQueue` maps its 404 (which can only mean the pool) to `ErrWorkPoolNotFound`, and the controller sets a `WorkPoolNotFound` condition with a 5s requeue, mirroring the automation controller's `DeploymentNotFound` — tested with the pool appearing later.

**The rest:** client unit tests for `workqueue.go` added (the 404 contracts, escaping, and payload shape including cleared-field defaults); printcolumns now show the queue name and concurrency limit; the `remote` nil-safety at the status write is restructured so both err-free paths guarantee it, with a comment; the `workQueueMatches` comment now states the removed-field consequence (handled via the clear list, not silently); `.DS_Store` removed and gitignored; `PROJECT`, `README.md`, and `deploy/samples/v1_prefectworkqueue.yaml` added.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	if queue.Priority != nil {
+		payload["priority"] = *queue.Priority
+	}
+	for _, f := range clearFields {
+		switch f {
+		case WorkQueueFieldConcurrencyLimit:
+			payload["concurrency_limit"] = nil
+		case WorkQueueFieldDescription:
+			payload["description"] = nil
```
chrisguidry: `payload["description"] = nil` will fail against a real server. `WorkQueue.description` is NOT NULL:

```
name: nullable=False        description: nullable=False
is_paused: nullable=False   concurrency_limit: nullable=True
priority: nullable=False
```

`WorkQueueUpdate.description` is `Optional[str]`, so Pydantic accepts the null; `model_dump_for_orm(exclude_unset=True)` keeps it because `exclude_none` defaults to false; and `models/workers.py:824-828` writes it with `.values(update_values)`. That hits the constraint.

The create-time default is `""` (`actions.py:964-966`, `orm_models.py:1129`), which is what the doc comment above says this sends. Use `payload["description"] = ""`.

The other two clears are right: `concurrency_limit` is nullable, and `is_paused` correctly sends `false`.

internal/prefect/workqueue_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+		It("sends explicit defaults for cleared fields", func() {
+			mockServer = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				defer GinkgoRecover()
+				var body map[string]any
+				Expect(json.NewDecoder(r.Body).Decode(&body)).To(Succeed())
+				// Explicit null clears the limit and description; is_paused is a
+				// plain bool server-side, so its default is an explicit false.
+				Expect(body).To(HaveKeyWithValue("concurrency_limit", BeNil()))
+				Expect(body).To(HaveKeyWithValue("description", BeNil()))
```
chrisguidry: This locks in the null. Once the clear sends the create-time default, it becomes `HaveKeyWithValue("description", "")`. The mock models description as nullable too, so the controller specs cannot catch this either.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+		if err == nil {
+			log.Info("Created work queue", "workQueue", workQueue.Name, "prefectId", remote.ID)
+		}
+	} else if err == nil {
+		// The queue already exists — created implicitly by a deployment
+		// referencing it, or previously managed. Adopt it rather than fail;
+		// only the first sync decides adoption, so a queue we created and then
+		// re-read stays "created".
+		if workQueue.Status.Adopted == nil {
```
chrisguidry: `status.adopted` is decided once and never revisited, but the managed queue is now keyed by `(workPoolName, name)` and `spec.name` is mutable. A rename changes which queue this resource manages, so adoption has to be recomputed.

Created `a` with `adopted=false`, then rename to `b` which already exists: the operator adopts `b`, `adopted` stays false, and deleting the CR deletes a queue the operator never created. That is the case the flag was added to prevent.

Adopted `a` with `adopted=true`, then rename to `b` which does not exist: the operator creates `b`, `adopted` stays true, and deleting the CR leaves `b` behind for good.

Recording the managed name in status (`status.managedName`) and recomputing `adopted` when `spec.name` differs from it covers both. `status.appliedFields` needs the same reset: renaming onto an existing queue while dropping a field applies the clear to a queue this resource never set that field on.

chrisguidry (commented):
Re-reviewed against `prefect` at `3.8.1.dev3`. The test suite passes locally.

Everything from the last round is addressed, and the CRD field docs are much better. The route move is correct: all five methods now use the pool-scoped family, so priority renormalization and the default-queue guard apply. Thanks for splitting the `client.go` escaping out into #313.

Two problems in the new code, both in parts that changed.

One cross-reference worth having here, since #314 applies the same pattern: the explicit null is correct there and wrong here. A deployment's concurrency limit lives in a related `ConcurrencyLimitV2` row that gets deleted (`models/deployments.py:445-450`), while a work queue's `description` is a NOT NULL column on the queue row itself. Same mechanism, different storage, opposite answer.

internal/controller/prefectworkqueue_controller.go
```diff
@@ -0,0 +1,358 @@
...
+	var clears []string
+	for _, f := range workQueue.Status.AppliedFields {
+		if f != prefect.WorkQueueFieldPriority && !slices.Contains(declared, f) {
+			clears = append(clears, f)
+		}
+	}
+
+	if err == nil && (len(clears) > 0 || !workQueueMatches(workQueue.Spec, remote)) {
+		err = prefectClient.UpdateWorkQueue(ctx, workQueue.Spec.WorkPoolName, workQueue.Spec.Name, spec, clears)
+		if errors.Is(err, prefect.ErrWorkQueueNotFound) {
```
chrisguidry, line 205: This returns without a status update, where the earlier version set a `Recreating` condition. It does recover, since `needsSync` is still true on the next pass either way. In between, the resource reports `Ready: true` with a stale `Synced` condition.

internal/controller/prefectworkqueue_controller_test.go
```diff
@@ -0,0 +1,440 @@
...
+			Expect(result.RequeueAfter).To(BeJitteredResync(testResyncInterval))
+
+			By("A queue exists under the new name")
+			renamed, err := mockClient.GetWorkQueue(ctx, "example-pool", "ingest-v2")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(renamed).NotTo(BeNil())
+			Expect(renamed.ConcurrencyLimit).To(HaveValue(Equal(int32(1))))
+
+			By("The old queue was not renamed, deleted, or modified")
+			old, err := mockClient.GetWorkQueue(ctx, "example-pool", "ingest")
```
chrisguidry, line 349: The rename spec does not assert `status.adopted`. Worth covering both directions: rename from a created queue onto a name that already exists, and rename from an adopted queue onto a name that does not.

internal/prefect/workqueue.go
```diff
@@ -0,0 +1,276 @@
...
+		return fmt.Errorf("work queue %q in pool %q: %w", name, workPoolName, ErrWorkQueueNotFound)
+	}
+	if !isSuccessStatusCode(resp.StatusCode) {
+		return fmt.Errorf("API request failed with status %d: %s", resp.StatusCode, string(body))
+	}
+	return nil
+}
+
+// DeleteWorkQueue deletes a work queue via DELETE /work_pools/{pool}/queues/{name},
+// which rejects deleting the pool's default queue cleanly and renormalizes the
```
chrisguidry: "rejects deleting the pool's default queue cleanly" overstates it. `models/workers.py:918-921` raises a bare `ValueError`, and the handler at `api/workers.py:1051-1076` does not catch it, so it surfaces as a 500 rather than a 4xx. The operator logs and proceeds, so the behavior is right; only the wording needs a change.

In practice this path is nearly unreachable now, since a pool's default queue always pre-exists and is therefore always adopted.

internal/prefect/workqueue.go
ShmuelOps: Fixed — the clear sends `""` now (and the doc comment names the per-field defaults: null limit, `""` description, `false` is_paused). Good catch on the storage difference vs #314.

internal/prefect/workqueue_test.go
ShmuelOps: Updated to `HaveKeyWithValue("description", "")`, and the mock now models the NOT NULL column by clearing to `""` as well.

internal/controller/prefectworkqueue_controller.go
ShmuelOps: Implemented as suggested: `status.managedName` records the queue the last sync managed; when `spec.name` differs, adoption is recomputed for the new queue and pending clears are dropped (they were applied against the old queue). Deletion also cleans up `managedName` rather than an unsynced `spec.name`.

internal/controller/prefectworkqueue_controller.go
ShmuelOps: Restored — the PATCH-404 race path sets the `Recreating` condition and `Ready: false` before the requeue again.

internal/controller/prefectworkqueue_controller_test.go
ShmuelOps: Added both directions: created→renamed onto an existing queue (re-adopts, the dropped field is not cleared on a queue that never had it, CR deletion leaves it) and adopted→renamed onto a new name (marked created, CR deletion removes it, the originally adopted queue untouched).

internal/prefect/workqueue.go
ShmuelOps: Reworded — it now says the rejection surfaces as a 500 (unmapped ValueError). Matches what I saw against a live 3.6.x server too.

chrisguidry (approved):

PR #314, opened 2026-07-28
fix: clear a removed deployment concurrency limit instead of silently keeping it

ShmuelOps:
## Summary

Removing `concurrencyLimit` from a `PrefectDeployment` was a silent no-op: the spec hash changes and a sync runs, but the field is neither compared (`concurrencyLimitMatches` skips a nil desired value) nor sent (`omitempty` + the server's `exclude_unset` update semantics), so the old limit stays in force while the resource reports `Ready: true`. We hit this in production use — the same gap @chrisguidry flagged for work queues in #311.

## What this does

The server distinguishes an **absent** `concurrency_limit` (leave unchanged) from an **explicit null** (remove the limit — `models/deployments.py` pops it against an `"unset"` sentinel). So, same mechanism as the #311 rework:

- `status.appliedFields` records which clear-tracked fields the last sync declared (currently `concurrencyLimit`; the list is the extension point for other optional fields with this hazard).
- A field present in `appliedFields` but gone from the spec is sent as an explicit `"concurrency_limit": null` via `UpdateDeployment`'s new `clearFields` argument.
- The up-to-date skip that protects auto-scheduled runs stays intact: `DeploymentClearsApplied` only forces an update while the remote still has the limit (read from `global_concurrency_limit.limit`; the top-level response field is deprecated and always null).

Verified against a live Prefect 3.6 server that `PATCH /deployments/{id}` with `"concurrency_limit": null` removes the limit (`global_concurrency_limit` returns null afterwards).

## Testing

- Controller spec: apply a limit, remove it, verify the remote limit is gone and `status.appliedFields` empties.
- Client specs: unset limit omitted from the payload; cleared limit sent as explicit null.
- Mock now models the server's `exclude_unset` semantics for the limit instead of full-replace, so the controller test actually exercises the bug.
- Full suite passes; `golangci-lint` clean; CRDs/deepcopy/docs regenerated.


internal/controller/prefectdeployment_controller.go
```diff
@@ -330,6 +342,16 @@ func scheduleNeedsUpdate(current, desired prefect.DeploymentSchedule) bool {
 }
 
 // setCondition sets a condition on the deployment status
+// declaredDeploymentFields lists the clear-tracked spec fields currently set,
```
chrisguidry: `declaredDeploymentFields` landed between the `setCondition` doc comment and `func setCondition`, so it now carries a comment about `setCondition` and `setCondition` has none. Move the new function below `setCondition`, or move the comment back down with it.

chrisguidry (commented):
Cross-referenced against `prefect` at `3.8.1.dev3`. Builds clean, full suite passes locally.

The mechanism checks out. `models/deployments.py:321` pops `concurrency_limit` against the `"unset"` sentinel and `:445-450` deletes the related limit row on an explicit null, so absent and null really are distinct. Reading `global_concurrency_limit.limit` is correct, given `responses.py:433-436` marks the top-level field as always None. `DeploymentClearsApplied` is doing real work: with the limit removed `DeploymentUpToDate` returns true, so #309's skip would otherwise swallow the clear, and the controller spec covers that path.

One thing to fix.

chrisguidry (approved):

internal/controller/prefectdeployment_controller.go
ShmuelOps: Fixed — moved `declaredDeploymentFields` below `setCondition` so the doc comment sits with its function again.

PR #316, opened 2026-07-29
fix: skip no-op automation updates so proactive trigger state survives resyncs

ShmuelOps:
Every `PUT /automations/{id}` resets the trigger's server-assigned ID — even when the payload carries the existing one — and proactive-trigger state buckets are keyed by `(automation_id, trigger_id)`. So each periodic resync silently discarded the trigger's accumulated state, and a proactive automation whose `within` exceeds the resync interval could never fire. Hit in production with the documented [zombie-flow detection](https://docs.prefect.io/v3/advanced/detect-zombie-flows) recipe (`within: 540s` vs 5-min resync): managed by the operator it never fired; the same automation created via the UI worked.

Three changes, mirroring the PrefectDeployment fix in #309:

- `AutomationUpToDate`: GET the automation on the update path and skip the PUT when every operator-managed field already matches (JSON-normalized subset compare; remote-only fields like trigger IDs and server defaults are ignored).
- `PreserveTriggerIDs`: carry existing trigger IDs onto the payload when a real update runs (top-level and compound/sequence children, same trigger type only) — effective on servers that honor IDs in updates.
- Only send `source` for action types whose server model has one; for the rest (e.g. `change-flow-run-state`) the server drops the key on write, which kept the stored action permanently different from the spec and defeated the skip.

Same caveat as #309: a field removed from the spec isn't detected by the subset compare until some other change forces an update (automations have no `status.appliedFields` yet, cf. #314).

Verified in a live deployment: with the fix, resyncs log "already up to date, skipping update", a single trigger ID accumulates buckets, and the zombie automation fires again.


chrisguidry (commented):
Two non-blocking notes. The core fix is right, and the skip is worth merging on its own.

### `PreserveTriggerIDs` has no effect on any Prefect server

`PUT /automations/{id}` calls `automation.trigger.reset_ids()` as its first statement, before validation or persistence ([`src/prefect/server/api/automations.py:102`](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/api/automations.py#L102); line 49 does the same for `POST`). It assigns a fresh UUID and recurses into the child triggers. I checked Prefect Cloud as well: its create and update paths reset trigger IDs the same way.

So no server honors a client-supplied trigger ID, and the payload change has no effect. I'd suggest dropping `PreserveTriggerIDs`, `toMapSlice`, and their three tests.

The reason is not just dead weight. The function and its comment tell a later reader that the trigger identity survives an update, which is the opposite of the behavior this PR works around. The skip is what protects the proactive state, and it does that job by itself.

### Minor: an unset `spec.enabled` gives inconsistent behavior

`Enabled` is a `*bool` with no CRD default, and `AutomationUpToDate` skips the field when it is nil. On the server, `AutomationUpdate.enabled` defaults to `True` and the model is `extra="forbid"`, so a `PUT` that omits the key re-enables the automation.

The result is that an automation paused in the UI stays paused across a resync, then gets re-enabled the moment any other field drifts and forces a real update. Before this PR, every resync re-enabled it. Defaulting `enabled` to `true` in the CRD would make the compare and the payload agree.

chrisguidry:
One more non-blocking note, on test coverage.

I hardwired `AutomationUpToDate` to return `true` and ran the full controller suite. It stayed green. So no controller test proves that a changed spec still reaches Prefect.

The new "drift resync fires with no changes" test covers the skip. The existing "When the spec changes" test does not cover the other side, because it asserts only `Ready` and a stable `Status.Id`, and both of those hold whether the update ran or not.

That matters because of the failure mode. If the compare regresses later, the operator silently stops applying spec changes. Every `PrefectAutomation` then drifts from its manifest while reporting `Ready: true` with no error and no log line. That is harder to notice than the bug this PR fixes.

`UpdateAutomationCalls` is already in place, so the assertion is two lines in the test that exists:

```go
Expect(mockClient.UpdateAutomationCalls).To(Equal(1))
```

plus a check that the stored automation carries the new description, which also proves the payload arrived rather than just the call.

internal/prefect/automation_diff.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := json.Unmarshal(b, &out); err != nil {
+		return v
+	}
+	return out
+}
+
+// PreserveTriggerIDs copies the server-assigned trigger IDs from the remote
+// automation onto the desired payload before an update, so a real spec change
+// keeps the trigger's identity. IDs only carry over while the type is unchanged.
+func PreserveTriggerIDs(desired, remote map[string]any) {
```
chrisguidry: This has no effect on any Prefect server.

`PUT /automations/{id}` calls `automation.trigger.reset_ids()` as its first statement, before validation or persistence ([`src/prefect/server/api/automations.py:102`](https://github.com/PrefectHQ/prefect/blob/main/src/prefect/server/api/automations.py#L102); line 49 does the same for `POST`). It assigns a fresh UUID and recurses into the child triggers. I checked Prefect Cloud as well: its create and update paths reset trigger IDs the same way.

I'd suggest dropping this function, `toMapSlice`, the three `TestPreserveTriggerIDs` cases, and the call site in `syncWithPrefect`.

The reason is not just dead weight. The function and its comment tell a later reader that the trigger identity survives an update, which is the opposite of the behavior this PR works around. The skip is what protects the proactive state, and it does that job by itself.

Non-blocking.

internal/prefect/automation_diff.go
```diff
@@ -0,0 +1,141 @@
...
+// resets the trigger's ID and therefore its proactive state) can be skipped.
+// Remote-only fields (trigger IDs, server defaults) are ignored.
+func AutomationUpToDate(remote *Automation, desired *AutomationSpec) bool {
+	if remote == nil || desired == nil {
+		return false
+	}
+	if remote.Name != desired.Name || remote.Description != desired.Description {
+		return false
+	}
+	if desired.Enabled != nil && *desired.Enabled != remote.Enabled {
```
chrisguidry, line 19: Minor: this makes an unset `spec.enabled` behave inconsistently.

`Enabled` is a `*bool` with no CRD default, so this check is skipped when it is nil. On the server, `AutomationUpdate.enabled` defaults to `True` and the model is `extra="forbid"`, so a `PUT` that omits the key re-enables the automation.

An automation paused in the UI therefore stays paused across a resync, then gets re-enabled the moment any other field drifts and forces a real update. Before this PR, every resync re-enabled it. Defaulting `enabled` to `true` in the CRD would make the compare and the payload agree.

Non-blocking.

internal/controller/prefectautomation_controller_test.go
```diff
@@ -174,6 +174,29 @@ var _ = Describe("PrefectAutomation controller", func() {
...
+			old := metav1.NewTime(time.Now().Add(-time.Hour))
+			fresh.Status.LastSyncTime = &old
+			Expect(k8sClient.Status().Update(ctx, fresh)).To(Succeed())
+
+			result, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: name})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result.RequeueAfter).To(BeJitteredResync(testResyncInterval))
+
+			// The remote already matches, so no update happened.
+			Expect(mockClient.UpdateAutomationCalls).To(BeZero())
```
chrisguidry, line 191: This covers the skip. Nothing covers the other side.

I hardwired `AutomationUpToDate` to return `true` and ran the full controller suite. It stayed green. The existing "When the spec changes" test asserts only `Ready` and a stable `Status.Id`, and both of those hold whether the update ran or not.

The failure mode is what makes it worth two lines. If the compare regresses later, the operator silently stops applying spec changes, and every `PrefectAutomation` drifts from its manifest while reporting `Ready: true` with no error and no log line.

`UpdateAutomationCalls` is already in place, so add this to that test:

```go
Expect(mockClient.UpdateAutomationCalls).To(Equal(1))
```

plus a check that the stored automation carries the new description, which proves the payload arrived and not just the call.

Non-blocking.

chrisguidry (commented):
Same three notes as above, now anchored to the code.

chrisguidry (commented):
Thanks again!  I do think there are a few minor things to address here before merging, but generally looking great!

chrisguidry (approved):

